PR #4, opened 2022-03-10
Fixed clippy issues

Eugeny:


mjc:
✅ Addressed Copilot feedback (commit 48fc6eb): Added `write_packet` API to `PacketWriter` that reuses `compress_buffer` as the payload scratch space, eliminating per-packet `Vec<u8>` allocations in `data_noqueue_direct`. Updated both callsites to use the new method.

PR #18, opened 2022-08-16
Allow vendoring openssl

connor4312:
This adds a `vendored-openssl` feature that will have the crate use the vendored version of OpenSSL rather than dynamically linking it.


Eugeny:
@allcontributors please add @connor4312 for code

Eugeny:
@all-contributors please add @connor4312 for code

PR #19, opened 2022-08-16
initial support bidirectional, server-initiated channels

connor4312:
Channels need not be exclusively client-created, in fact few (if any?)
channels specify directionality, and refer only to a sender and receiver.

This does an initial, somewhat minimal pass at making servers capable of
initiating channels. This is rather rough and I think greater
abstraction is warranted here; I pulled out relevant parsing code into
`parsing.rs`, but there is logical duplication between the client and
server which I don't have time to disentagle. This also only implements
_some_ channel functionality on the server-side, it's not as complete as
the client.

The significant breaking change here is that `server::run_stream` now
returns (almost) immediatel with a SessionHandle, similar to the client;
this is used to open channels.

Ultimately this is a big set of changes I would understand if there's
not great desire to merge this upstream :)

### DIFF
--- a/russh/src/channels.rs
+++ b/russh/src/channels.rs
@@ -1,0 +1,335 @@
+use russh_cryptovec::CryptoVec;
+use tokio::sync::mpsc::{Sender, UnboundedReceiver};
+
+use crate::{ChannelId, Error, Pty, Sig};
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ChannelMsg {
+    Open {
+        id: ChannelId,
+        max_packet_size: u32,
+        window_size: u32,
+    },
+    Data {
+        data: CryptoVec,
+    },
+    ExtendedData {
+        data: CryptoVec,
+        ext: u32,
+    },
+    Eof,
+    /// (client only)
+    RequestPty {
+        want_reply: bool,
+        term: String,
+        col_width: u32,
+        row_height: u32,
+        pix_width: u32,
+        pix_height: u32,
+        terminal_modes: Vec<(Pty, u32)>,
+    },
+    /// (client only)
+    RequestShell {
+        want_reply: bool,
+    },
+    /// (client only)
+    Exec {
+        want_reply: bool,
+        command: String,
+    },
+    /// (client only)
+    Signal {
+        signal: Sig,
+    },
+    /// (client only)
+    RequestSubsystem {
+        want_reply: bool,
+        name: String,
+    },
+    /// (client only)
+    RequestX11 {
+        want_reply: bool,
+        single_connection: bool,
+        x11_authentication_protocol: String,
+        x11_authentication_cookie: String,
+        x11_screen_number: u32,
+    },
+    /// (client only)
+    SetEnv {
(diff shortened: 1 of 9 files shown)


connor4312:
Oh, also -- after implementing the bulk of this, I realized that I didn't actually need it for our particular use case (I thought we needed to implement direct-tcpip from the server, but it was actually forward-tcpip). So I won't be heartbroken if this doesn't make the cut.

Eugeny:
Will need some time to wrap my head around it - stay tuned :)

BTW what's the use case for russh in vscode?

connor4312:
A couple months ago we released the beta of the [VS Code server](https://code.visualstudio.com/docs/remote/vscode-server). This is powered by the [forwarding service](https://github.com/microsoft/dev-tunnels) being built by a partner team in Azure (public release in a little while) which operates via SSH over websockets.

Since the SDKs are already public and my contribution to add tunneling will be OSS soon, I'll just share the docblock comment that explains our flow 🙂 

> Tunneling communicates via SSH over websockets. Data is sent fairly opaquely in
> binary websocket messages. We use Tungstenite for the websocket, and then
> wrap it into an AsyncRead/AsyncWrite type that we can give to the websocket
> library, russh.
>
> We then connect over SSH. Today, Tunneling doesn't require additional auth in
> the SSH connection, and also has a somewhat non-standard "none" value for
> its key exchange. This prevents many off-the-shelf SSH implementations from
> working. Once a client connects to the other end of the basis socket, it'll
> open a new Channel from the server (of type "client-ssh-session-stream").
>
> The stream of data over this channel is actually an SSH client connection.
> So, for each client, we create an SSH server instance, and do some more
> work to make that channel AsyncRead/AsyncWrite as well. (Note that this
> client actually does do a key exchange and its data is E2E encrypted.) Once
> established, the host (this program) requests the client to forward the
> ports that it via `tcpip-forward` requests. Clients then open
> `forwarded-tcpip` channels for each new connection.
>
> ```text
>        ┌───────────┐     ┌───────┐      ┌───────┐
>        │Host (this)│     │Service│      │Client │
>        └─────┬─────┘     └───┬───┘      └───┬───┘
>              │ Connect as    │              │
>              ├─SSH client────▶              │
>              │               │  Connect to  │
>              │               ◀──service ws──┤
>              │   Create new  │              │
>              ◀───SSH tunnel──┤              │
>              │               │              │
>              │    SSH server handshake.     │
>              ├────(Service just proxies ────▶
>              │      traffic through)        │
>              │               │              │
>              ├────tcpip-forward for ports───▶
>              │               │              │
>              │               │              ◀───asked to
>              │               │              │   connect
>              ◀────create forwarded-tcpip ───┤
>      make    │            channel           │
> local tcp ◀──┤               │              │
> connection   │               │              │
>              ◀ ─ ─ ─ ─forward traffic─ ─ ─ ─▶
>              │               │              │
>              │               │              │
>              ▼               ▼              ▼
> ```

Eugeny:
Wow, glad to be a part of it! :) I've merged the changes manually and added and tested forward-tcpip from the client side of things.

PR #20, opened 2022-08-16
add support forward-tcpip requests from the server

connor4312:
This is simple enough -- allows the server to ask for forwarded ports,
for which the client can create channels and then forward.

### DIFF
--- a/russh/src/lib.rs
+++ b/russh/src/lib.rs
@@ -320,6 +320,7 @@ macro_rules! push_packet {
     }};
 }
 
+mod parsing;
 mod session;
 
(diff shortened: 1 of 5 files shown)


Eugeny:
Is this specific to your use case? (asking since RFC says clients should reject `tcpip-forward` messages)

connor4312:
Ah, yea, I didn't notice that in the RFC previously--I suppose these are somewhat use case specific. Though the RFC uses "SHOULD" and justifies it due to it being "normally only sent by the client," so there's no hard restriction against the inversion.

PR #22, opened 2022-08-19
Add documentation for client

mihirsamdarshi:
hey this is a small contribution, but one I thought I would add since I myself was confused when using this crate. let me know if it needs anything.


Eugeny:
Looking great, much appreciated! @all-contributors please add @mihirsamdarshi for docs

PR #26, opened 2022-08-23
Silently handle openssh keepalive requests in client

amtelekom:
Using russh to connect to an openssh sshd server with a mostly idle connection, the log is filled with `info`-level events about `keepalive@openssh.com` channel request not being handled.
This works, since OpenSSH interprets the channel_failure just fine as a keepalive response, but fills the log with unnecessary noise.

Handling this non-standard, but widely used extension prevents that.


Eugeny:
Thank you, much appreciated! @all-contributors please add @amtelekom for code

PR #28, opened 2022-08-23
Some code clean up, made some code DRY

mihirsamdarshi:
In this PR I got rid of some "*" imports, and made some code DRY. I have been working a lot with the client code, and I have a future PR coming that should handle some failures that I see when opening a `direct-tcpip` channel

In addition I saw that keep alive request handling was added in #26, however, these requests happen via a Global Request message rather than Channel Request message, so I fixed that behavior.

Let me know if this is acceptable


Eugeny:
Looks great, thanks a lot!

PR #30, opened 2022-08-24
remove unnecessary mut markers on the server Handle

connor4312:
The handle is Sync, so removing mutability allows it to be used in an Arc. A little more efficient than cloning the handle around.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -64,7 +64,7 @@ pub struct Handle {
 
 impl Handle {
     /// Send data to the session referenced by this handler.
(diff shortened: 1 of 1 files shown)


Eugeny:
The reason for `&mut self` is to allow the trait implementors to mutate themselves (especially important for `Handler` since it's very likely to have complex behaviour). Unless I'm missing something, it's not absolutely possible to mutate self from a trait method that only has `&self`

connor4312:
This is on the Handle itself, not the Handler trait

Eugeny:
duh, my bad!

Issue #34, opened 2022-09-05
Support custom server_id line terminators.


evilsocket:
Hi, I opened this [same issue on the original thrussh repo](https://nest.pijul.com/pijul/thrussh/discussions/51) but it was never fixed/implemented, trying again here before I go creating my own fork just for this small change.

I’m using thrussh to create an SSH server honeypot. Among the configuration options, the user can specify the server_id (which is basically what nmap and most of the tools use to fingerprint the SSH server). The problem is that some banners are terminated by a single “\n”, while others are terminated by “\r\n”, but the library hardcodes the latter here https://github.com/warp-tech/russh/blob/master/russh/src/sshbuffer.rs#L40-L44

If someone passes a banner like “something\n”, that breaks the protocol because “\r\n” is added anyway. You can find more details about this issue here https://github.com/evilsocket/medusa/issues/3

I suggest to add the server_id termination bytes as a configurable option (default to “\r\n”). Let me know if you prefer to do the changes yourself or you’d prefer me sending a PR ^_^

Eugeny:
Hi, the RFC says that `\r\n` is explicitly required and only some older undocumented implementations used `\n`. I won't be able to find time to spend on such quirks, but will accept a PR changing the type of `server_id` to something along the lines of

```rust
enum ServerID {
  Normal { software: String, version: String },
  Raw(String)
}
```

PR #36, opened 2022-09-05
new: support for custom server_id line terminators (fixes warp-tech/russh#34)

evilsocket:
I tried to keep it as simple as possible.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -560,9 +560,10 @@ impl Session {
                             originator_port,
                         } => {
                             confirm();
+                            let channel = self.accept_server_initiated_channel(id, &msg);
                             client
                                 .server_channel_open_x11(
-                                    id,
+                                    channel,
                                     originator_address,
                                     *originator_port,
                                     self,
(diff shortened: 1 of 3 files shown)


Eugeny:
The value used for kex hash generation should exclude the linebreaks.

evilsocket:
ops, my bad, fixing

evilsocket:
this creates a few issues in case of a custom (raw) server_id which includes line breaks ... I could trim them forcefully from the string but I don't like this solution ... this change should only affect send_ssh_id. Maybe it's better to just extend it as:

```rust
    pub fn send_ssh_id(&mut self, id: &[u8], custom_line_breaks: Option<&[u8]>) {
        self.buffer.extend(id);
        match custom_line_breaks {
            Some(bytes) => self.buffer.extend(bytes),
            None => {
                self.buffer.push(b'\r');
                self.buffer.push(b'\n');
            }
        }
    }
```  

And then add those bytes as a configuration (default to None so it falls back to the standard default). 

What do you think?

evilsocket:
sent a new commit fixing the issue

Eugeny:
I've taken it upon myself to clean it up a little bit and merged :v:

Eugeny:
@all-contributors please add @evilsocket for code

evilsocket:
thanks a lot! looking forward to the changes :D

Eugeny:
Should be released on crates.io now

evilsocket:
i was referring to the server_id changes :D

PR #40, opened 2022-09-07
implement clone for keypair()

connor4312:
In my case, I want to make multiple servers with the same keypair, so this allows it to be cloned.


Eugeny:
Pls check the lint output :v:

PR #41, opened 2022-09-07
decrease log level verbosity for ordinary messages

connor4312:
Currently the "debug" level is filled with lots of verbose info about individual messages. Imo the "debug" level should contain information about state changes, while the "trace" level contains information about flowing data. That is, debug is should be useful to general consumers who want to know about connection details and reliability, while trace can be used for people who want to dig into the protocol itself :)


Eugeny:
Good idea, agreed!

PR #43, opened 2022-09-26
Also handle channel keepalives

amtelekom:
Hi,

my OpenSSH 8.7p1 sends keepalives via channels, not global requests, as soon as I have multiple channels open.

This also matches with a [short look at the openssh source](https://github.com/openssh/openssh-portable/blob/master/serverloop.c#L149).

Also, write keepalives to the log at a lower loglevel, a keepalive interaction isn't noteworthy of an INFO log entry IMHO.


Eugeny:
LGTM 🚀

PR #48, opened 2022-10-09
Server-side channel handles

Eugeny:


AspectUnk:
You've got to be kidding me :_D Interesting coincidence.

PR #49, opened 2022-10-09
AsyncRead/AsyncWrite channels

Eugeny:


connor4312:
Btw, I did this in https://github.com/microsoft/dev-tunnels/blob/b20ef08c719f843a1d9ba8610c52071fae6a482f/rs/src/connections/relay_tunnel_host.rs#L938 too. You might find that helpful/inspirational 🤷‍♂️

Eugeny:
that's exactly what I wanted to make, but good :D

PR #52, opened 2022-10-12
add flags for openssl-only crypto

### LINKED ISSUE #50: FIPS "OpenSSL only" mode
One cloud on the horizon for our usage of this library is the likelihood of needing to be [FIPS](https://csrc.nist.gov/publications/detail/fips/140/2/final) certifiable. Long story short, this essentially means having all cryptography done by a dynamically linked OpenSSL library--which in turn is [FIPS certified](https://csrc.nist.gov/Projects/cryptographic-module-validation-program/validated-modules). In this case, we would use OpenSSL for the cryptography currently done by crates like `aes` and `pbkdf2`.

I will (attempt to) implement this in our fork of the library, but I was wondering if it's something there'd be interest in merging upstream as a feature flag. Understanding that dependency on OpenSSL is not greatly desirable and that this would increase the complexity of the library for something a minority of consumers will ever need 😛

connor4312:
Fixes #50

This introduces an on-by-default `rs-crypto` flag, which enables the existing Rust-based crypto libraries (including aes and ED25519). However, these implementations can be removed by disabling the flag. If it's disabled, then openssl (when turned on) will stand in for them, in a less performant way.

Note that while OpenSSL 3.x does have some ED25519 support, I have not done the work to make that compatible as well--partly because ED25519 is not yet an approved algorithm for my company to use, and partly to retain compatibility with OpenSSL 1.x


connor4312:
I will fix up the build/tests in the morning, it's late here already (but it works!)

Eugeny:
:+1: merge c1472f6 for CI setup

connor4312:
Hmm, not sure what's up with the build failure; that test works on my machine -- though later doctests fail since a lot of them show examples with ed25519 keys... Not sure whether I'll have time to dive into that more today or not

Eugeny:
No worries and no rush

PR #55, opened 2022-10-18
Add ability to close a channel from the client

jgrund:
This patch addresses two issues:
1. A client should be able to send a `close` message to the server via a channel
2. When the client receives the CHANNEL_CLOSE message it must also respond to the server wit hthe same message per the spec (see https://datatracker.ietf.org/doc/html/rfc4254#section-5.3)

> When either party wishes to terminate the channel, it sends
> SSH_MSG_CHANNEL_CLOSE.  Upon receiving this message, a party MUST
> send back an SSH_MSG_CHANNEL_CLOSE unless it has already sent this
> message for the channel.  The channel is considered closed for a
> party when it has both sent and received SSH_MSG_CHANNEL_CLOSE, and
> the party may then reuse the channel number.  A party MAY send
> SSH_MSG_CHANNEL_CLOSE without having sent or received
> SSH_MSG_CHANNEL_EOF.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -328,6 +328,9 @@ impl Session {
                 let mut r = buf.reader(1);
                 let channel_num = ChannelId(r.read_u32().map_err(crate::Error::from)?);
                 if let Some(ref mut enc) = self.common.encrypted {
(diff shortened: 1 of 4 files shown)


Eugeny:
Much appreciated! @all-contributors please add @jgrund for code

PR #56, opened 2022-10-18
Implement detached signal sender

jgrund:
Allow signals to be sent detached from the channel struct.

This is done using a higher-order function and by cloning the sender so it can be used independently of the `Channel`.

This is useful to be able to spawn a cancelation handler separately from the main `Channel` recieve loop.

Signed-off-by: Joe Grund <jgrund@whamcloud.io>


Eugeny:
Isn't this something that can be solved with `tokio::select`? (for any commands, not just `.signal()`)

```rust
let channel = ...
let (tx, rx) = mpsc::channel(1);

loop {
    tokio::select! {
        msg = channel.wait() => ...
        sig = rx.recv() => channel.signal(sig)
    }
}

...

tx.send(Sig::...)
```

jgrund:
> Isn't this something that can be solved with tokio::select? (for any commands, not just .signal())

Not really, because I do not want to hold  a `Channel` ref, only a way to send signals to it.

My usecase is a map of signal senders that I can use to interrupt open connections on demand (orchestration platform).

Eugeny:
You don't need to hold the ref for that - the ref can stay in the `wait()` loop, while you can move `tx` where you need it and use it to send signals

jgrund:
>  the ref can stay in the wait() loop

Yeah, this is the ref I was talking about. 

It's honestly easier to have the channel return me a (scoped) sender directly than to have to create a layer of indirection that does the same thing.

Eugeny:
Can't merge this as it's a band-aid fix for one specific method that will eventually lead to duplication of every channel method. 

However I'm open to adding a `Channel.handle` method that would return a handle object that's similar to Channel but only has "sender write" methods such as data/signal/close etc. Could even be a trait shared between it and Channel to avoid duplication.

Let me know if this is something you'd like to work on, otherwise I'll see if I can find time for it in the future.

PR #87, opened 2022-11-30
Replace SocketAddr to ToSocketAddrs trait

AspectUnk:
Using ToSocketAddrs when specifying SocketAddr would be more appropriate, correct and convenient. This is the approach used in net libs


Eugeny:
LGTM, thank you! @all-contributors please add @AspectUnk for code

PR #91, opened 2022-12-07
Adding cfg flags for the tests

AspectUnk:
The structure and the test that uses UnixStream is for some reason not marked as unix only, which leads to an error on windows. UnixStream is only available for unix.


Eugeny:
Thanks!

PR #95, opened 2022-12-10
Added test for ChannelStream & fix tests for windows

AspectUnk:
ChannelStream works well for me. Unfortunately I don't have my macbook right now to run tests, but on windows I still noticed errors that I had to fix.


Eugeny:
What a coincidence, I've _just_ added the same test :D 

Can you see if maybe you could add your changes to it? Currently it only checks ChannelStream in a client session

Eugeny:
Also you can actually replace `whoami` with just `"user".to_string()` since the tests just test against `russh` itself and not a real SSH server

AspectUnk:
:D
Okay, I'll make changes soon.

PR #115, opened 2023-01-30
don't panic when passphrase is not supplied to openssh private key

mariotaku:
When passphrase is empty for an OpenSSH bcrypt key, russh will panic. This PR makes it returns in the result instead.


Eugeny:
Thank you for the contribution! @all-contributors please add @mariotaku for code

Issue #109, opened 2023-01-09
Help Wanted: Using Russh to execute multiple commands


mTsBucy1:
For https://github.com/mTsBucy1/async-ssh2-tokio-fork I would like to be able to sequentially execute many commands on the same host (See also https://github.com/Miyoshi-Ryota/async-ssh2-tokio/issues/12).

Currently we create one Handler Implementation and call `connect` to get a Session Handle and authenticate this.
We then call `channel_open_session()` and `channel.exec` for every command that needs execution.
This was done, because calling `channel.exec` multiple times has failed.

However, after exactly 10 invocations, the session gets disconnected (presumably by the server). This was demonstrated on a Windows OpenSSH server and a Ubuntu ssh server.

I've tried looking into the russh code to understand the issue, but without success. The one thing I tried, was to override `Handler::channel_close` to also send a `session.close(channel);` packet back to the server. This was unsuccessful, but I also found no reference of a channel ever getting closed any other way.

Is this an inherit limitation of the session and do we need to reconnect and/or reauthenticate every 10 invocations? Or is there something we missed?
If someone with more SSH knowledge could point me in the right direction that would be amazing.
If you need a minimal reproducable example or further explanations from my side, please ask, but I maybe someone here knows of a approach to solve this.

Eugeny:
This is actually expected. You can't execute multiple commands on a single channel, and in most distros, OpenSSH server config limits the max number of channels, both simultaneous and total. Make sure that you're closing channels after the command is complete so that you're not hitting the lower simultaneous channel limit

mTsBucy1:
Thanks for your answer @Eugeny, I was just wondering how I can close the channel manually. I couldn't find any function in the documentation, neither on the Handle nor on the Channel. As stated, I tried to override `Handler::channel_close` to send a Close message back, because I definitely get a Close message from the server, but firstly, it didn't increase the limit, and secondly, I now have a memory leak, because I can't remove the channel from the private HashMap holding the channels in the Session.

And just to clarify, the ssh connection will regularly have to be reconnected (or only re-authenticated?) every once in a while?

Eugeny:
You're correct! For some reason there is no `Channel` method to close it yet - so channels currently can only be closed via [Session::close](https://docs.rs/russh/0.35.0-beta.9/russh/client/struct.Session.html#method.close)

Eugeny:
Should be fixed now.

mTsBucy1:
Thanks for the quick fix 🚀

[...]

PR #118, opened 2023-02-01
Try to fix channel close issue

yorkz1994:
This related to my comment on https://github.com/warp-tech/russh/issues/109#issuecomment-1410006553.

I managed to find the real problem.
In russh/src/client/encrypted.rs:
image
When received channel close message, it removes the channel first from channel map, then send back channel_close message to server by ```Encrypted::byte``` function.
But in ```Encrypted::byte``` function, it checks the channel map to actually send the message, since it has been removed, the close message never got send to the server, this keeps the server channel leak.

image

My solution is to move remove channel in map code into ```Encrypted::close``` function.  

1. If client initiate close, ```Encrypted::close``` will be invoked, close message will be sent to server and removed in local channel map, if later received server close message, as channel has been removed, no close message will be send again.

2. If server initiate close, ```Encrypted::close``` will be invoked too, close message will be sent to server and removed in local channel map.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -340,9 +340,7 @@ impl Session {
                 if let Some(ref mut enc) = self.common.encrypted {
                     // The CHANNEL_CLOSE message must be sent to the server at this point or the session
                     // will not be released.
-                    if enc.channels.remove(&channel_num).is_some() {
-                        enc.byte(channel_num, msg::CHANNEL_CLOSE);
-                    }
+                    enc.close(channel_num);
                 }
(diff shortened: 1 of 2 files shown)


Eugeny:
Wow - nice find, thank you for the contribution! @all-contributors please add @yorkz1994 for code

yorkz1994:
@Eugeny,

Do you plan for a new release to crates.io for this fix?

Eugeny:
Just released v0.36.1 :v:

yorkz1994:
Excellent!
Thanks!

Issue #121, opened 2023-02-08
`ssh-rsa` partially supported in `russh-keys` (both `AgentClient` and `Signature`)


cipriancraciun:
If the user creates a `russh_keys::PublicKey::parse ("ssh-rsa", bytes)` (as opposed to getting one from the agent, or using `"rsa-sha2-{256,512}"`), and tries to use that with the current `russh_keys::agent::AgentClient`, the code fails to decode the `ssh-rsa` signature.

cipriancraciun:
Thanks for merging the pull request.

This issue is now solved, closing.

PR #122, opened 2023-02-08
Add support for `ssh-rsa` signatures in `russh-keys`

cipriancraciun:
As described in #121 the current client code fails to correctly parse `ssh-rsa` signatures.

This minor patch just adds that support to parse these signatures.


Eugeny:
Looking great! :+1: 

@all-contributors please add @cipriancraciun for code

PR #128, opened 2023-02-18
Use shared refs instead of mutable refs in client Handle, where possible

mllken:
Some client Handle operations can take a long time to finish (e.g. channel_open_direct_tcpip of a down hostname), so holding a mutable reference can lock the Handle.  This change switches Handle's self to shared references, where possible.


Eugeny:
Could you please explain how it locks the handle? Are you trying to perform multiple operations on the `Handle` at the same time?

mllken:
Yes, when I say lock, I mean I have to use a Arc+Mutex with a call to lock() when using the handle concurrently.  With this change, I can get rid of the Mutex, when using the handle concurrently.

Eugeny:
Thanks. @all-contributors add @mllken for code

PR #130, opened 2023-02-18
Send empty chunk on AsyncWrite#shutdown for `ChannelStream`

simao:
This is needed so that the closure in `into_stream` knows that it needs to send a close message down the channel when `AsyncWrite#shutdown` is called. Otherwise the channel always stays open.

Additionally, change `poll_read` in `AsyncRead` for `ChannelStream` to return `Ok(())` on EOF, when it doesn't write to `buf`. This is the contract for `AsyncRead` as documented in
https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html#tymethod.poll_read Otherwise the future always terminates with an error.


Eugeny:
Thank you!

Issue #108, opened 2023-01-09
Direct Streamlocal 


lmtr0:
Hi there, is there any plan to support direct stream local ?

Eugeny:
It's supported on client side now, server side: #323

PR #131, opened 2023-02-18
Initial client support for unix socket (direct-streamlocal) channel connections

mllken:
Allow clients to initiate unix socket connections.  The method naming follows the direct_tcpip naming convention.  Updates https://github.com/warp-tech/russh/issues/108


sybernomad:
Is there a reason why the `self.channel_open_generic` method in sessions.rs in both the client and server are private?

Eugeny:
`channel_open_generic` requires the caller to manually fill in the payload, and `russh` aims to provide a safe interface with no access to underlying SSH messages.

PR #132, opened 2023-02-18
Call session handler disconnected when session disconnects.

simao:
I am not sure this should be called somewhere else?

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -463,7 +463,10 @@ impl Session {
                 .map_err(crate::Error::from)?;
             self.common.write_buffer.buffer.clear();
         }
+
         debug!("disconnected");
+        handler.disconnected(self).await?;
+
         // Shutdown
         stream_write.shutdown().await.map_err(crate::Error::from)?;
         loop {


Eugeny:
It looks like this method was added mistakenly in #19. The "normal" way to know when the session ends is to implement `Drop` for your handler - russh will drop it as soon as the session disconnects.

simao:
I see. That would kind of work but the problem is on drop we don't have access to `Session`, which we have on all the other callbacks. So if you need to implement some cleanup logic it would be good to have access to session, in particular `Session#id`, so we could correlate a Server Handler to a Session. 

I guess this could be worked around.. setting some field on the handler in one of the other callbacks and then getting that field on drop, to correlate a session to a Handler, but this feels like a hack compared to the other callbacks? Are we guaranteed that a Handler instance is only used for a single Session?

Eugeny:
Yes, a `Handler` is linked to its `Session` (you're creating a unique Handler each time in [`Server::new_client`](https://docs.rs/russh/latest/russh/server/trait.Server.html#tymethod.new_client)), and they're destroyed together too. So, by the time the disconnect happens (`Handler::drop` or otherwise), the`Session` itself is already dead and there's nothing you can do with it apart from calling that `id()` method.

I recommend handling session identification yourself, e.g. with

```rust
struct Handler {
    id: usize,
}

impl server::Handler for Handler { ... }
impl Drop for Handler { ... }

struct Server { last_id: usize  }

impl server::Server for Server {
    type Handler = Handler;
    
    fn new_client(&mut self, _) -> Handler {
        self.id += 1;
        Handler { id: self.last_id }
    }
}


```

simao:
Ok, thank you!

PR #133, opened 2023-02-18
remove all "extern crate" references and other minor changes

mllken:
- Update cryptovec's Rust edition to 2018, to match all the other russh crates.
- Remove all references to 'extern crate', which was a relic from Rust 2015.
- Remove the call to `#![recursion_limit = 128]`, as that is already the default (https://doc.rust-lang.org/reference/attributes/limits.html)

I've rebuilt and ran all tests, so hopefully this is OK.


Eugeny:
LGTM 🚀

PR #134, opened 2023-02-18
Replace serde-derive crate with serde + the derive feature

mllken:
The serde_derive crate is no longer needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,7 +1243,6 @@ dependencies = [
  "rand_core 0.5.1",
  "russh-cryptovec",
  "serde",
- "serde_derive",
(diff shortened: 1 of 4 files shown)


Eugeny:
Hey, this error like means you've have changes in Cargo.lock that you didn't commit :v:

mllken:
Ah, OK, I have now added it.  I think for libraries, the best practice is usually to not check in Cargo.lock, so I had purposely skipped it. https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

Eugeny:
Good point - it's already in .gitignore but somehow made its way into the repo anyway - deleted!

PR #144, opened 2023-03-03
fix #143: handle CHANNEL_OPEN_FAILURE on server side

### LINKED ISSUE #143: Server: Handle::channel_open_forwarded_tcpip stuck when client fails to connect to remote host
**Description**

Hi, first of all thank you for this great project.
While implementing remote port forwarding (`ssh -R`), I got this issue: if client fails to connect to the designated service, it replies to server with `CHANNEL_OPEN_FAILURE` (92u8).

My issue is that I'm waiting for this reply to continue the process. But it seems that the server drops it.

In the server logs, I found this: `src/server/encrypted.rs:897: unknown message received: Some(92)`.
I assume this is the packet I'm waiting for with `Handle::channel_open_forwarded_tcpip`.

**Environment**

- russh: "0.36.1" (from crates.io)
- logs gathered using `tracing-log`

Please tell me if more details are required. Thanks by advance for your help.

Swelio:
This a small fix to handle CHANNEL_OPEN_FAILURE messages on server side. It fixed my issue so I propose the solution.
Please feel free to correct and add anything I could have missed.

### DIFF
--- a/russh/src/server/encrypted.rs
+++ b/russh/src/server/encrypted.rs
@@ -893,6 +893,23 @@ impl Session {
                     }
                 }
             }
+            Some(&msg::CHANNEL_OPEN_FAILURE) => {
+                debug!("Channel open failed.");
+                let mut buf_pos = buf.reader(1);
+                let channel_num = ChannelId(buf_pos.read_u32().map_err(crate::Error::from)?);
+                let reason =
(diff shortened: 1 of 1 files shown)


Eugeny:
Looks good to me, thank you so much!

@all-contributors add @Swelio for code

PR #146, opened 2023-03-04
correct initial fix for issue 143 (channel open failure on remote port forwarding)

Swelio:
Hello. Thank you for your reply and merge, happy to help!
I realized that my initial solution didn't address the problem correctly. It was only replying to client with the same error code.

This new patch send `ChannelMsg::OpenFailure` as a response to `Handle::wait_channel_confirmation` which can then return an error as `Err(Error::ChannelOpenFailure(reason))` to `Handle::channel_open_forwarded_tcpip`.

This has effectively fixed my own project and the method is not stuck anymore. Sorry for the mistake, I'm quite new on this project so I will take time to read the code in the future.

### DIFF
--- a/russh/src/server/encrypted.rs
+++ b/russh/src/server/encrypted.rs
@@ -895,7 +895,7 @@ impl Session {
                 }
             }
             Some(&msg::CHANNEL_OPEN_FAILURE) => {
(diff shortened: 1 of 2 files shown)


Eugeny:
No worries and thanks for the fix!

PR #148, opened 2023-03-07
Handle tcp-ip-forward global request responses.

jimmyvanhest:
When a tcp-ip-forward global request is sent a REQUEST_SUCCESS or a REQUEST_FAILURE response is expected and should be handled. this is also an opportunity to parse and return the port number returned by the server.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -559,6 +559,20 @@ impl Session {
                 }
                 Ok((client, self))
             }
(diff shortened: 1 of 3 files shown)


Eugeny:
Welcome and thank you for your contribution!

Unfortunately `read_u32` is failing when there's no port info in the request (e.g. in the success response of the `cancel-tcpip-forward` request)

The session should either remember the type of the last global request and act accordingly or in the worst case, just ignore a missing port field.

PR #158, opened 2023-05-16
support RFC8731 name of curve25519-sha256 kex

jcgruenhage:
Since RFC8731 got published in February 2023, this kex method has been
officially called `curve25519-sha256`, and the `@libssh.org` suffix is really not
necessary anymore. `openssh` has been treating both the same since December 2016.


Eugeny:
Thanks! Please also add `CURVE25519_PRE_RFC_8731` to `Preferred::DEFAULT`

PR #161, opened 2023-06-19
Error::UnsupportedKeyType holds a String

rukai:
Previously errors looked like: `UnsupportedKeyType([114, 115, 97])`
But with this PR they now look like: `UnsupportedKeyType("rsa")`

### DIFF
--- a/russh-keys/src/format/mod.rs
+++ b/russh-keys/src/format/mod.rs
@@ -76,7 +76,10 @@ pub fn decode_secret_key(secret: &str, password: Option<&str>) -> Result<key::Ke
             } else if l == "-----BEGIN RSA PRIVATE KEY-----" {
                 #[cfg(not(feature = "openssl"))]
                 {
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you for your contribution @rukai , however `russh` aims to provide an unambiguous interface to the SSH protocol. Conditionally stringifying `key_type` here creates a situation where it's unclear whether a string was sent as-is or is a result of `format!()`

rukai:
Some alternatives:
* format as escape characters with https://doc.rust-lang.org/std/ascii/fn.escape_default.html
* format as "string: {}" or "bytes: {}", depending on if its utf8 or not.
* have both a string and raw bytes `UnsupportedKeyType { string: String, raw: Vec<u8> }`

Would one of these work?
I'm sure we can find a solution where the error is both readable and unambiguous.

Eugeny:
Making `UnsupportedKeyType` a struct variant would be fine by me :)

rukai:
Alright, I rewrote the PR to have both a raw and string field in a struct variant.

PR #166, opened 2023-07-25
Trait method to add conditions for SSH agent server when accepting requests for operations 

### LINKED ISSUE #165: Allow creating custom conditions for SSH agent server accepting requests for certain keys
I've been writing [an ssh-agent replacement](https://github.com/gotlougit/sshield) for fun lately, and I was using this crate to do so. However, I've been stumped on creating a prompt for every request that the SSH agent gets, and from the look at the code responsible for handling this, there doesn't seem to be an obvious way to do this from outside the crate.

Maybe I'm missing something, but could this be something added to be implemented in the `russh_keys::agent::server::Agent` trait that could (optionally) be implemented by a crate user?

gotlougit:
Fix #165 

The `russh_keys::agent::server::Agent` trait now has another async trait we can define if necessary which will check if we should be able to perform a given operation or not. External developers can implement this trait method with their own code, allowing the request to fail if necessary.

The operation types are represented in a newly created `MessageType` enum and passed into this new method. By default, the method returns true, so old code shouldn't be affected by this change.

### DIFF
--- a/russh-keys/Cargo.toml
+++ b/russh-keys/Cargo.toml
@@ -29,6 +29,7 @@ version = "0.37.1"
 
 [dependencies]
 aes = "0.8"
+async-trait = "0.1.72"
 bcrypt-pbkdf = "0.9"
 bit-vec = "0.6"
 cbc = "0.1"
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you for the PR! Could you please check the `.expect()` usage? As russh doesn't allow panics, you need to return an `Err` in that case, otherwise LGTM :v:

Eugeny:
Perfect! @all-contributors please add @gotlougit for code

PR #169, opened 2023-08-02
Update dependencies

rukai:
Notes:
* It seems appropriate to just unwrap the result of `pbkdf2` since I assume it was previously just panicking internally on such a case.
* bitflags 2.0 now automatically implements IntoIterator, removing the need for manually implementing Iterator. It also requires us to Derive a bunch of traits ourselves that used to be done automatically.
* a change in curve25519-dalek resulted in a previous double reference failing to automatically dereference. Not sure why that is but changing it to a single reference makes more sense anyway.
* I left the rand in russh-keys as is because that version of rand is used in the ed25519-dalek api.

### DIFF
--- a/russh-keys/Cargo.toml
+++ b/russh-keys/Cargo.toml
@@ -30,14 +30,14 @@ version = "0.37.1"
 [dependencies]
 aes = "0.8"
 async-trait = "0.1.72"
-bcrypt-pbkdf = "0.9"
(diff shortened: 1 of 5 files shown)


rukai:
alright, I backed out the `pbkdf2` changes since I'm not sure how to handle that properly, it can be done by someone else.

PR #174, opened 2023-08-17
Use negotiated KEX instead of preferred

### LINKED ISSUE #16: Curve25519 kex panic
I replied below [this issue](https://github.com/warp-tech/russh/issues/15), but since the issue has been closed,I am worried that no one will see it, so I opened this issue.

RDruon:
The KEX algo was always the first of the preferred KEX one.

Instead relies on the KEX algo that was negotiated with server rather than the preferred one.

Fix #16


Eugeny:
Ohhh great find! :+1: 
Took me a bit to notice the error.

PR #180, opened 2023-09-12
Use `ChannelMsg::WindowAdjusted` during data transfer

### LINKED ISSUE #179: many `ChannelMsg::WindowAdjusted` messages buffered during large data transfer
Continuation of #101 

Hi,

I'm writing a large amount of data (~ 4G) to a remote OpenSSH server running `scp -t`.

Along the way, the OpenSSH server starts sending `CHANNEL_WINDOW_ADJUST` messages. 

We only attempt to read a single `ChannelMsg::WindowAdjusted` message when our `channel.window_size` reaches 0. This leaves a large number of messages in the unbounded channel.

When my `channel.data` call returns, I drain the channel with a `channel.wait`, and count the ChannelMsg::WindowAdjusted that are sitting in the channel. I see around 40_000 unhandled `ChannelMsg::WindowAdjusted` that get drained. 

It seems like we should either: 

- Check for `ChannelMsg::WindowAdjusted` in the channel and adjust before we reach 0 window_size

- Remove `ChannelMsg::WindowAdjusted` as a message type and perhaps create a separate https://docs.rs/tokio/latest/tokio/sync/index.html#watch-channel for it.

Thoughts?

jgrund:
OpenSSH server sends `CHANNEL_WINDOW_ADJUST` messages before window_size is 0.

Handle these messages at each turn of the loop within `Channel.send_data`

Fixes #179

### DIFF
--- a/russh/src/channels.rs
+++ b/russh/src/channels.rs
@@ -1,6 +1,6 @@
-use russh_cryptovec::CryptoVec;
-use tokio::sync::mpsc::{Sender, UnboundedReceiver};
 use log::debug;
+use russh_cryptovec::CryptoVec;
+use tokio::sync::mpsc::{error::TryRecvError, Sender, UnboundedReceiver};
 
 use crate::{ChannelId, ChannelOpenFailure, ChannelStream, Error, Pty, Sig};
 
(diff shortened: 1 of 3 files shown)


Eugeny:
Just a heads-up - I've reverted your PR in favour of #181. There's nothing wrong with your PR - just that #181 includes more robust handling in case of multiple parallel writers and other improvements :v:

Issue #99, opened 2022-12-17
client channel extended data AsyncReader


mllken:
Hi,

I'm really liking the new async trait handlers!

I can turn a channel into a AsyncRead + AsyncWrite stream using Channel's into_stream() method.  Could you also provide a method to stream a Channel's stderr?  Possibly, it could be a AsyncReader for extended data for the channel, special cased for where the extended_code == 1.  

I'm using russh-0.35.0-beta.8
Thanks

Eugeny:
Now available in v0.39.0-beta.1

Issue #162, opened 2023-07-05
How to serve basic shell request using russh


michaelfortunato:
Hi all,
I am trying to write a very basic ssh daemon. Essentially what it does is
1) Listens for clients
2) For each client, forks a process (where the program is inter-active (std::process:Command("/bin/bash") for instance))
3) Connects the stdin and stdout of the forked process to the client.

Here is my approach so far. I will 
1) Override server::Handler::channel_open_session
2) Allocate a pty in that function
3) Launch the child process
4) Set the child process's stdin and stdout to be the pty slave file.
3) On a tokio thread copy the master file to the channel
4) On another tokio thread, copy the channel to the master file

```rust
use std::collections::HashMap;
use std::process::Command;
use std::sync::Arc;
use std::time::Duration;
use std::{os, process, thread};

use async_trait::async_trait;
use pty::fork::Fork;
use russh::server::Auth;
use russh::server::{Msg, Session};
use russh::*;
use russh_keys::*;

impl server::Handler for Server<'_> {

    async fn channel_open_session(
        self,
        channel: Channel<Msg>,
        session: Session,
    ) -> Result<(Self, bool, Session), Self::Error> {
        let fork = Fork::from_ptmx().unwrap();
        if let Some(mut master) = fork.is_parent().ok() {
            tokio::spawn(async {
                std::io::copy(&mut channel, &mut master); // This does not work because channel does not implement Read
            });
            tokio::spawn(async {
                std::io::copy(&mut master, channel); // This does not work because channel does not implement Write
            });

        } else {
            let Some(mut child) = fork.is_child().ok()
            Command::new("/bin/bash").stdin(child).stdout(child).spawn();
        }
        Ok((self, true, session))
    }
```

The above code won't compile because channel does not implement the Read and Write functions for io::copy. 

My questions are
1) Is my approach correct in general for the problem I am trying to solve (ie. allocate an interactive process for each client)
2) If my approach makes sense, is there a struct in russh that implements Read and Write?

Eugeny:
Your approach in general is ok. `Channel` doesn't implement Read/Write, but you can turn it into a `tokio::AsyncRead + AsyncWrite` via [`Channel::into_stream`](https://docs.rs/russh/latest/russh/struct.Channel.html#method.into_stream)

You can't use `std::io::copy` with it though, and will have to write a copy loop yourself using `tokio::select!`, since the `ChannelStream` can't be split into Read and Write halves.

michaelfortunato:
[...]

Issue #178, opened 2023-09-09
channel stream keeps returning zero bytes even if read to end 


Irvingouj:
```rust 
use anyhow::Result;
use async_trait::async_trait;
use russh::*;
use russh_keys::*;
use std::sync::Arc;
use tokio::io::{AsyncReadExt, AsyncWriteExt};

struct Client {}

#[async_trait]
impl client::Handler for Client {
    type Error = anyhow::Error;

    async fn check_server_key(
        self,
        server_public_key: &key::PublicKey,
    ) -> Result<(Self, bool), Self::Error> {
        println!("check_server_key: {:?}", server_public_key);
        Ok((self, true))
    }

    async fn data(
        self,
        channel: ChannelId,
        data: &[u8],
        session: client::Session,
    ) -> Result<(Self, client::Session), Self::Error> {
        println!(
            "data on channel {:?}: {:?}",
            channel,
            std::str::from_utf8(data)
        );
        Ok((self, session))
    }
}

#[tokio::main]
async fn main() -> Result<()> {
    simple_logger::SimpleLogger::new().init()?;
    let config = russh::client::Config::default();
    let config = Arc::new(config);
    let sh = Client {};

    // let mut agent = russh_keys::agent::client::AgentClient::connect_env().await?;
    // agent.add_identity(&key, &[]).await?;
    let mut session = russh::client::connect(config, ("192.168.50.194", 22), sh).await?;
    session
        .authenticate_password("irving", "990225ojy")
        .await
        .unwrap();

    let mut channel = session
        .channel_open_direct_tcpip("www.example.com", 80, "127.0.0.1", 51241)
        .await
        .unwrap();

    let mut stream = channel.into_stream();
    let request = b"GET / HTTP/1.1\r\nHost: www.example.com\r\nConnection: close\r\n\r\n";
    stream.write(request).await.unwrap();

    let mut buf = Vec::<u8>::new();
    let mut counter = 0;
    loop {
        if let Ok(bytes_read) = stream.read_to_end(&mut buf).await {
            println!(
                "msg red = {:?}, ascii = {:?}",
                &buf[..bytes_read],
                String::from_utf8_lossy(&buf[..bytes_read])
            );
            buf.clear();
            counter +=1;
            if counter >=10{
                panic!()
            }
        } else {
            break;
        }
    }

    Ok(())
}
```

if we run this code, we will get this log at the end.
```
mains/example\">More information...</a></p>\n</div>\n</body>\n</html>\n"
[...]

PR #181, opened 2023-09-12
Implement Channel::into_io_parts() for AsyncRead and AsyncWrite data transfers.

Nurrl:
Hi,

as mentioned in #162, I wanted to contribute to this crate by adding an easy way to pipe a `tokio::process::Command` I/O to a SSH Channel.

This PR introduces a few changes:
- A structure `ChannelTx`, implementing `AsyncWrite`
- A structure `ChannelRx`, implementing `AsyncRead`
- Two new methods on the `Channel` struct:
```rust
/// Setup the [`Channel`] to be able to send and receive [`ChannelMsg::Data`]
/// through [`io::ChannelTx`] and [`io::ChannelRx`].
pub fn into_io_parts(&mut self) -> (io::ChannelTx<S>, io::ChannelRx<'_, S>) {
    self.into_io_parts_ext(None)
}

/// Setup the [`Channel`] to be able to send and receive [`ChannelMsg::Data`]
/// or [`ChannelMsg::ExtendedData`] through [`io::ChannelTx`] and [`io::ChannelRx`]
/// depending on the `ext` parameter.
pub fn into_io_parts_ext(
    &mut self,
    ext: Option<u32>,
) {
    /* ... */
}
```
This implementation allows to Drop either of the structs independently and regain access to the `Channel` struct when dropping the `Rx` side.
- As a minor change, I also did create a directory for the `channels` module to house the original code as well as my new `io` module.
- ~~It optionally includes an _auto-close mechanism for the SSH Channels_, it improves a bit the handling of `Channel`s but might be a **breaking** change~~: **Edit: Reverted due to breaking-change and failing a test.**

<details>
  <summary>Code snippet</summary>

```rust
impl<S> Drop for Channel<S>
where
    S: From<(ChannelId, ChannelMsg)>,
{
    fn drop(&mut self) {
        let _ = futures::executor::block_on(
            self.sender
                .send((self.id, ChannelMsg::Close).into())
                .inspect_err(|err| {
                    log::error!("Failed to automatically close channel #{}: {err}", self.id)
                }),
        );
    }
}
```

</details>

---

Finally the usage of this is as such:
```rust
pub async fn execute(
    mut command: tokio::process::Command,
    channel: &mut russh::Channel<russh::server::Msg>,
) -> eyre::Result<std::process::ExitStatus> {
    let mut child = command.spawn()?;

    let (mut stdout, mut stdin) = (
        child
            .stdout
            .take()
            .expect("Unable to take service's `stdout` handle"),
        child
            .stdin
            .take()
            .expect("Unable to take service's `stdin` handle"),
    );
    let (mut tx, mut rx) = channel.into_io_parts();

    tokio::try_join!(
        async move {
            let res = tokio::io::copy(&mut rx, &mut stdin).await;
            drop(stdin); // Dropping stdin hints the process about the tty being disconnected

            res
        },
        tokio::io::copy(&mut stdout, &mut tx)
    )?;

    Ok(child.wait().await?)
}
```

Thanks for reading,
Feel free to ask questions, or add suggestions :)


Nurrl:
Upon reviewing the design and some issues here and there, I figured the issue with splitting `rx` and `tx` in the current design is the handling of the `ChannelMsg::WindowAdjusted` messages for data transfer, so I also made a proposal to move the handling of those messages in the `Session::run` loop to alleviate the need to manage it at multiple places.

See 166d71e6baa2be3e517f0d1efa9edf2b061d6311.

Eugeny:
Appreciate your work and the direction is looking good on the first look. I'm away until mid next week but should be able to properly review after that :v:

Nurrl:
Thanks for your quick response, have a great week until then :tada:

Eugeny:
Just checked it and everything looks good except a few nitpicks on the API.

* You've removed `Channel::writable_packet_size` - can you readd it back to `ChannelTx`?
* The `into_io_parts` is a bit wonky since it invites comparison with `Socket::into_split`, but doesn't actually consume the `Channel` as the `into_` prefix implies. Moreover, the way to creating two `Tx` sides for stdout and stderr (calling `into_io_parts` twice on the same object) is not obvious and feels wrong to the user. I suggest splitting it into `make_writer` and `make_reader`, which makes it clear that the halves are independent and can be created separately and/or multiple times.

Nurrl:
Hi,
thanks for your review, I plan to do the changes tomorrow.

- Sure, I'll add the `writable_packet_size` method again.
- Agreed on the `into_io_parts` method, about the `make_reader` and `make_writer`, do you wish them to take the `ext: Option<u32>` parameter directly or have separate counterparts for extended data (`make_reader_ext` and `make_writer_ext`) ?
- About the `ChannelStream` wouldn't it be better to drop it altogether, since both ends are independent from each-other and can close at different moments ? this would be a breaking change though. 
- Also, do you want me to remove the two commits about the close-on-drop from history as mentioned in the original comment ? (The commit itself and it's revert)

> can you please also add tests to cover the new code

For sure, however I'm unsure about how to create such test, could it work with an integration test with a client and a server sending data to each other within a `tokio::join` ? Is there an example of a similar test ?
I think unit-testing could be a bit hard because of the nature of the `WindowAdjusted` packets.

Thanks !

russh/src/channels/mod.rs
```diff
@@ -111,7 +118,7 @@ pub struct Channel<Send: From<(ChannelId, ChannelMsg)>> {
     pub(crate) sender: Sender<Send>,
     pub(crate) receiver: UnboundedReceiver<ChannelMsg>,
     pub(crate) max_packet_size: u32,
-    pub(crate) window_size: u32,
+    pub(crate) window_size: Arc<Mutex<u32>>,
```
jgrund, line 121:
```suggestion
    pub(crate) window_size: Arc<AtomicU32>,
```

Possible to do this instead to avoid `Mutex`?

Nurrl: Hi there, thanks for the suggestion, I though of using Atomics for the `window_size`, however there are two reasons this is not suitable for this case:
- Atomics operations might be blocking which is strongly discouraged in an _async_ context, because it might block the event loop and there are no non-blocking counterparts of those operations (ala `Mutex::try_lock`);
- We _must_ hold the Mutex in the following code to ensure there are no race-conditions because we first fetch and then update the value, this means no other thread can act on the `window_size` and lead to an overshoot with two threads sending data causing the window to be exceeded.
```rust
let mut window_size = match self.window_size.try_lock() {
    Ok(window_size) => window_size, 
    Err(_) => {
        cx.waker().wake_by_ref();
        return Poll::Pending;
    }
};

let writable = self.max_packet_size.min(*window_size).min(buf.len() as u32) as usize;
if writable == 0 {
    cx.waker().wake_by_ref();
    return Poll::Pending;
}

let mut data = CryptoVec::new_zeroed(writable);
#[allow(clippy::indexing_slicing)] // Clamped to maximum `buf.len()` with `.min`
data.copy_from_slice(&buf[..writable]);
data.resize(writable);

*window_size -= writable as u32;
drop(window_size);
```

Nurrl:
Kinda related topic, but is there a reason to take `&mut self` in the `Channel` methods now that the `window_size` is thread-safely handled ? Most (if not all) of these methods are only requiring a `&self` reference.

Eugeny:
Nope, feel free to change them to &self where possible :v:

Nurrl:
It's done :tada:, about my proposition to remove the `ChannelStream` altogether [here](https://github.com/warp-tech/russh/pull/181#issuecomment-1728372919), what do you think ?

Once this is sorted the last thing to do will be the tests :)

Eugeny:
I think we should keep the `ChannelStream` interface (even if as a simple wrapper) since it's used by e.g. `russh_sftp` which expects an `AsyncRead + AsyncWrite`

Nurrl:
I have some troubles sorting the `ChannelStream` matter because of `io::ChannelRx`'s lifetime, is there a `russh` chat channel somewhere I can join and discuss design ?

Eugeny:
I don't have a channel but I guess we can just discuss it in this thread :) I see two options:

* Add a lifetime to `ChannelStream`
* Change `ChannelRx` to consume `Channel` instead of referencing it: `Channel::into_reader(self) -> ChannelRx` + `ChannelRx::into_channel(self) -> Channel`

Nurrl:
Okay, I wasn't sure if it was okay to flood here.

- Adding the lifetime to `ChannelStream` is what I went for, but unfortunately, `SftpSession::new` requires `S: 'static`, I could bubble down this change if you agree.
- I'd say changing `ChannelRx` to consume `Channel` would mean we would have no way to close the channel once we created a `ChannelRx` which is what I wanted to prevent when I set it to capture `&mut Channel`.

The other way this could be done would be with the help of `owning_ref` but it seems unmaintained and to have some RUSTSEC issues..

Nurrl:
I'm not really fond of the `into_reader()` + `into_channel()` design because it requires owning the channel beforehand and is a pain to work with when it's part of a structure.

Nurrl:
I ended up creating an `enum` that can either hold an owned struct or a ref of Channel to solve the issue.

Nurrl:
Other thing, I feel like the `S: From<(ChannelId, ChannelMsg)` constraint could be removed altogether with the template type, since we always end up creating `Msg` in the Channel. Is there a reason this template is here ?

Eugeny:
@Nurrl it's due to there being two different `Msg` enums (for the server and the client). Nice trick with `ChannelAsMut`!

Nurrl:
Hi again, I fixed some flushing issues and added an end-to-end test for the data streaming from client to server and back.

Eugeny:
LGTM! @all-contributors please add @Nurrl for code

Nurrl:
Merging might have closed #99 & #178 @Eugeny :eyes:

Eugeny:
@Nurrl could you please also make `ChannelRx` and `ChannelTx` public? And make `ChannelStream::new` `pub(crate)`

Nurrl:
Hi there, just came back from week-end, sure for ChannelStream, even pub(super) works, about ChannelRx and ChannelTx, isn't it better to even hide them via a `impl` ? They don't expose any methods and keeping the public interface small helps mitigate breaking changes.

Eugeny:
Agreed!

PR #184, opened 2023-09-16
bumb toolchain & add sftp client example

AspectUnk:
Hi again ☺️ I added a client part to `russh-sftp` for interacting with the protocol at any level, including high-level no worse than most crates. Please add an example, but I saw that this causes errors because your crate uses an outdated toolchain `1.65.0`, although a stable `1.70.0` has already been released which adds a lot of new things, if this does not cause problems for you, update it.


Eugeny:
Thank you!

PR #195, opened 2023-10-15
Make winapi dep windows only

rukai:
This PR improves build time and disk usage by avoiding downloading the source for the winapi crate when not on windows.


Eugeny:
Thank you!

PR #196, opened 2023-10-27
Add client-sent keepalives

mmirate:
This is analogous to OpenSSH's `ServerAliveInterval` option and is of similar purpose. When the crate user is executing a command that will run for some time without producing output or expecting any input, it becomes difficult to distinguish between "the server has nothing to send" and "the connection to the server has been severed". A periodic keepalive puts an upper bound on how long an unexpected disconnection will go unnoticed in the absence of data traffic.


Eugeny:
Thank you for the contribution! 
@all-contributors please add @mmirate for code

PR #203, opened 2023-11-01
Fix handling of key constraints

george-hopkins:
I tested the ssh-agent client against OpenSSH and found some differences in the encoding of key constraints. This PR ensures both implementations are interoperable. In addition, the russh ssh-agent will actually store key constraints (it didn't until now). The client will check the response to report back any errors.


Eugeny:
Thank you!

PR #208, opened 2023-11-05
Support for NIST P-256 public keys

george-hopkins:
This PR implements support for `ecdsa-sha2-nistp256` public keys based on [`p256`](https://crates.io/crates/p256).


Eugeny:
Looking great, thank you!

Eugeny:
I've also started working on the key support in the client in #209 if you're interested in reviewing

PR #211, opened 2023-11-06
Client Timeout

amtelekom:
Implement timeouts for SSH clients during normal operation.

So far, this was only possible in servers.

Note that this doesn't include provisions for sending keepalives, so unless the server sends those faster than the chosen timeout, the application needs to track this itself for now.

### DIFF
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -882,9 +882,10 @@ impl Session {
             }
         }
         debug!("disconnected");
-        if self.common.disconnected {
-            stream_write.shutdown().await.map_err(crate::Error::from)?;
-        }
+        self.receiver.close();
+        self.inbound_channel_receiver.close();
+        stream_write.shutdown().await.map_err(crate::Error::from)?;
+
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you!

mmirate:
> Note that this doesn't include provisions for sending keepalives, so unless the server sends those faster than the chosen timeout, the application needs to track this itself for now.

@amtelekom, could you please explain what the difference between what you're looking for regarding keepalives, and https://github.com/warp-tech/russh/pull/211/files#diff-a54f5b4b2c7b6955dd496e3f9b7bcc52dac6226eb3450c4785bd7957e47f0c1dR782-R785? (as added in #196)

Eugeny:
@mmirate this PR is confusingly named - it adds an automatic disconnect if no server message (keepalive or others) is received within `inactivity_timeout`

mmirate:
@Eugeny Ah, so given that in #196 the client-sent keepalives are hardcoded to `want_reply = true`, then I suppose that the author meant to say: the application is responsible for ensuring that its provided configuration satisfies `keepalive_interval < inactivity_timeout`, plus round-trip-time plus processing time plus uncertainty margin?

PR #212, opened 2023-11-09
Fix ChannelMsg::close docs

rukai:
As per the protocol specification: https://datatracker.ietf.org/doc/html/rfc4254#section-5.3
`SSH_MSG_CHANNEL_CLOSE` can be sent by the client as well as the server, so the `(server only)` comment is incorrect.


Eugeny:
Thanks!

PR #214, opened 2023-11-13
Improve #196 and #211

mmirate:
* Rewrite these two features using disjunctive futures in order to cleanly make the timers optional.
  This works because of https://docs.rs/futures/latest/futures/future/enum.Either.html#impl-Future-for-Either%3CA,+B%3E.
  It theoretically works best because reusing the same futures means Tokio has to do less bookkeeping. Every non-diverging exit from the `select!` will reset the keepalive timer, and will also reset the inactivity timer if any nontrivial I/O happened.
* Add an analogue of OpenSSH's `ServerAliveCountMax`.
  A counter is persisted on the `Session`; incremented when a keepalive is sent; and reset when a REQUEST_SUCCESS or REQUEST_FAILURE is received (presumably in reply to a keepalive). If it passes a configurable threshold, the connection is cut.
* Mutate the `Session` to pass information back to the main bg loop from the plaintext packet reader, so that only nontrivial data transfer will reset the inactivity timer. (And so that `ServerAliveCountMax` will be judged correctly.)

### DIFF
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -790,7 +790,7 @@ impl Session {
         #[allow(clippy::panic)] // false positive in select! macro
         while !self.common.disconnected {
             self.common.received_data = false;
-            let mut only_sent_keepalive = false;
+            let mut sent_keepalive = false;
             tokio::select! {
                 r = &mut reading => {
                     let (stream_read, mut buffer, mut opening_cipher) = match r {
@@ -834,13 +834,13 @@ impl Session {
                     reading.set(start_reading(stream_read, buffer, opening_cipher));
                 }
                 () = &mut keepalive_timer => {
-                    self.send_keepalive(true);
-                    only_sent_keepalive = true;
                     if self.common.config.keepalive_max != 0 && self.common.alive_timeouts > self.common.config.keepalive_max {
                         debug!("Timeout, server not responding to keepalives");
                         break
                     }
                     self.common.alive_timeouts = self.common.alive_timeouts.saturating_add(1);
+                    self.send_keepalive(true);
+                    sent_keepalive = true;
                 }
                 () = &mut inactivity_timer => {
                     debug!("timeout");
@@ -899,7 +899,7 @@ impl Session {
                 }
             }
 
-            if self.common.received_data {
+            if self.common.received_data || sent_keepalive {
                 if let (futures::future::Either::Right(ref mut sleep), Some(d)) = (
                     keepalive_timer.as_mut().as_pin_mut(),
                     self.common.config.keepalive_interval,
@@ -907,7 +907,7 @@ impl Session {
                     sleep.as_mut().reset(tokio::time::Instant::now() + d);
                 }
             }
-            if !only_sent_keepalive {
+            if !sent_keepalive {
                 if let (futures::future::Either::Right(ref mut sleep), Some(d)) = (
                     inactivity_timer.as_mut().as_pin_mut(),
                     self.common.config.inactivity_timeout,
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -105,8 +105,8 @@ use crate::session::{CommonSession, EncryptedState, Exchange, Kex, KexDhDone, Ke
 use crate::ssh_read::SshRead;
 use crate::sshbuffer::{SSHBuffer, SshId};
 use crate::{
-    auth, msg, negotiation, strict_kex_violation, ChannelId, ChannelOpenFailure,
-    Disconnect, Limits, Sig,
+    auth, msg, negotiation, strict_kex_violation, ChannelId, ChannelOpenFailure, Disconnect,
+    Limits, Sig,
 };
 
 mod encrypted;
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -128,8 +128,6 @@ pub struct Session {
     pending_len: u32,
     inbound_channel_sender: Sender<Msg>,
     inbound_channel_receiver: Receiver<Msg>,
-    server_alive_timeouts: usize,
-    activity: bool,
 }
 
 const STRICT_KEX_MSG_ORDER: &[u8] = &[msg::KEXINIT, msg::KEX_ECDH_REPLY, msg::NEWKEYS];
@@ -700,6 +698,8 @@ where
             disconnected: false,
             buffer: CryptoVec::new(),
             strict_kex: false,
+            alive_timeouts: 0,
+            received_data: false,
         },
         session_receiver,
         session_sender,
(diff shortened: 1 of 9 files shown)


mmirate:
This is related to #196 and #211. @amtelekom, how badly did I misunderstand what you were looking for?

russh/src/client/mod.rs
```diff
@@ -723,6 +723,16 @@ async fn start_reading<R: AsyncRead + Unpin>(
     Ok((n, stream_read, buffer, cipher))
 }
 
+fn future_or_pending<F: futures::Future, T>(
```
Eugeny, line 726: Could this function be avoided by using `.is_some()` as a precondition in `tokio::select!`?

amtelekom:
> how badly did I misunderstand what you were looking for?

First of all thanks for looking out for my needs :)

Your branch works fine for our purposes - we just need to be able to detect if the network connectivity was lost or the Server exit without closing down the connection correctly on long-standing SSH connections (Netconf transport tunnels).

Since our (OpenSSH-based) servers can send keepalives, just copying the `inactivity_timeout` from the Russh server side was the quickest solution, but actually sending own keepalives and then timing out on non-responsiveness is actually the better solution to our specific situation.

While it works, it feels like the activity tracking seems like it goes farther than needed?
As mentioned, using the `inactivity_timeout` config just came from the Russh Server side, where it is just used to add a timeout to the `select!` loop.
In your current PR, the interpretation of the flag differ between the Server and Client side.

It might still be the better way of tracking "inactivity", but then maybe we should also port this to the Server side for consistency, and expand the explanation in the doc comment in the `Config` struct?

russh/src/client/mod.rs
mmirate: I'm afraid not. An `Option<impl Future>` doesn't also implement `Future`, and even if the precondition disables a branch, it still always evaluates the async expression and expects it to be the same type each time. So there has to be _something_ pollable to put in the place of an absent `Sleep`, and the `Pending` future is the obvious choice since it will efficiently never complete.

That said, this function has only two callsites, and could be inlined back into those callsites - I only factored it out for DRY purposes.

mmirate:
> First of all thanks for looking out for my needs :)
> 
> Your branch works fine for our purposes - we just need to be able to detect if the network connectivity was lost or the Server exit without closing down the connection correctly on long-standing SSH connections (Netconf transport tunnels).

Well, I would marvel at the coincidence that our respective employers have the same motivation for us to use russh, but I guess there aren't that many other client-side usecases of SSH that can't be done "simpler" by just shelling out to OpenSSH's client program with tokio::process. (Though I'd assume that that would be less scalable since certainly each russh Session has to be using less memory than each entire `ssh` process...)

> Since our (OpenSSH-based) servers can send keepalives, just copying the `inactivity_timeout` from the Russh server side was the quickest solution, but actually sending own keepalives and then timing out on non-responsiveness is actually the better solution to our specific situation.

Ah, so that feature's use case now makes a lot more sense.

> While it works, it feels like the activity tracking seems like it goes farther than needed?

My best guess is that what you meant is: the activity tracking should only say "even though a packet was read, it didn't merit resetting the inactivity timer" if the packet in-question was clearly a keepalive request or reply. Is this correct?

(By contrast, what I wrote will also avoid resetting the inactivity timer in case of XON/XOFF, unhandled global requests, unhandled channel requests and unhandled packets. Upon reflection, I see that's clearly a subjective judgment of mine.)

> As mentioned, using the `inactivity_timeout` config just came from the Russh Server side, where it is just used to add a timeout to the `select!` loop. In your current PR, the interpretation of the flag differ between the Server and Client side.
> 
> It might still be the better way of tracking "inactivity", but then maybe we should also port this to the Server side for consistency, and expand the explanation in the doc comment in the `Config` struct?

You're right, I have thusfar been paying zero attention to the Server side of russh. I'll take some time to see where this all fits into the Server.

mmirate:
> I'll take some time to see where this all fits into the Server.

I have concluded this process, and thus I believe my recent commit brings Client and Server into parity regarding these two features.

Based on what I saw on the Server side - and contrary to the code I wrote in my previous PR - it seems to me that that the purpose of the inactivity timer, relative to OpenSSH-style ServerAliveInterval or ClientAliveInterval, is that the inactivity timer on the one hand doesn't let the peer hold the connection open by responding to a ping-pong, but on the other hand does let the local application hold the connection open by sending data to the peer even without expecting any responses whatsoever to that data (not even window size adjustments or rekeyings or etc). So that is what I have assumed when writing this last commit.

amtelekom:
Hmm,

running your updated commit with

```rust
let config = Arc::new(sshclient::Config {
  keepalive_interval: Some(Duration::from_secs(60))
  ..sshclient::Config::default()
});
let client = SSHClient {};
let mut ssh = sshclient::connect(config, socket_addr, client).await?;
// ...
```

against a default-configuration OpenSSH Server (i.e. it doesn't send keepalives on its own) keeps runing into timeouts.

Activating the Debug outputs for russh, it seems like there is no wait time between the keepalive attempts being sent, and thereby also not giving the remote side more than a few ns to respond (the previous russh debugmsg timestamp before this is exactly a minute older):

```
2023-11-22T09:53:18Z DEBUG russh::cipher: writing, seqn = 558   
2023-11-22T09:53:18Z DEBUG russh::cipher: padding length 4   
2023-11-22T09:53:18Z DEBUG russh::cipher: packet_length 32   
2023-11-22T09:53:18Z DEBUG russh::cipher: writing, seqn = 559   
2023-11-22T09:53:18Z DEBUG russh::cipher: padding length 4   
2023-11-22T09:53:18Z DEBUG russh::cipher: packet_length 32   
2023-11-22T09:53:18Z DEBUG russh::cipher: writing, seqn = 560   
2023-11-22T09:53:18Z DEBUG russh::cipher: padding length 4   
2023-11-22T09:53:18Z DEBUG russh::cipher: packet_length 32   
2023-11-22T09:53:18Z DEBUG russh::cipher: writing, seqn = 561   
2023-11-22T09:53:18Z DEBUG russh::cipher: padding length 4   
2023-11-22T09:53:18Z DEBUG russh::cipher: packet_length 32   
2023-11-22T09:53:18Z DEBUG russh::client: Timeout, server not responding to keepalives   
2023-11-22T09:53:18Z DEBUG russh::client: disconnected   
2023-11-22T09:53:18Z DEBUG russh::client: drop session 
```

mmirate:
Aye, that was a nasty little oversight. Apologies. It should work correctly with that simple fix, and since your code was slightly cleaner than what I came up with prior to reading the entirety of your message, I cherry-picked your commit into here.

mmirate:
The semver checker failed when I rebased this onto the latest commit in `main`; I assume because the version numbers no longer have the prerelease indicator.

mmirate:
Heads-up, @Eugeny: [it looks like russh-sftp is using features that weren't stable in Rust 1.65](https://github.com/warp-tech/russh/actions/runs/7250114068/job/19749564779?pr=214#step:3:520).

Eugeny:
fixed!

mmirate:
Excellent! I took an educated guess at the version-bump, so CI seems to be in order now.

amtelekom:
Is there something that still needs to happen before this can be merged?

In case it helps: I feel this MR now implements things as I would have expected them to work from a user's perspective, and we've been working base on this branch successfully the last weeks. Would be good to be able to go back to the upstream version though :)

Issue #215, opened 2023-11-22
Example showing how to implement shell_request -> Command -> stdin/stdout/stderr?


brandonros:
```rust
use std::collections::HashMap;
use std::process::Stdio;
use std::sync::Arc;

use async_trait::async_trait;
use log::*;
use russh::server::{Auth, Msg, Session, Response};
use russh::*;
use russh_keys::*;
use tokio::io::{BufReader, AsyncReadExt, AsyncWriteExt};
use tokio::process::{Child, Command};
use tokio::sync::{mpsc, Mutex};

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    env_logger::init();
    let mut config = russh::server::Config::default();
    config.auth_rejection_time = std::time::Duration::from_secs(3);
    config
        .keys
        .push(russh_keys::key::KeyPair::generate_ed25519().unwrap());
    let config = Arc::new(config);
    let sh = Server {
        clients: Arc::new(Mutex::new(HashMap::new())),
        shells: Arc::new(Mutex::new(HashMap::new())),
        id: 0,
    };
    russh::server::run(config, ("0.0.0.0", 2222), sh).await?;
    Ok(())
}

#[derive(Clone)]
struct Server {
    clients: Arc<Mutex<HashMap<(usize, ChannelId), Channel<Msg>>>>,
    shells: Arc<Mutex<HashMap<ChannelId, Child>>>,
    id: usize,
}

impl server::Server for Server {
    type Handler = Self;
    fn new_client(&mut self, _: Option<std::net::SocketAddr>) -> Self {
        debug!("new client");
        let s = self.clone();
        self.id += 1;
        s
    }
}

#[async_trait]
impl server::Handler for Server {
    type Error = anyhow::Error;

    async fn channel_open_session(
        self,
        channel: Channel<Msg>,
        session: Session,
    ) -> Result<(Self, bool, Session), Self::Error> {
        {
            debug!("channel open session");
            let mut clients = self.clients.lock().await;
            clients.insert((self.id, channel.id()), channel);
        }
        Ok((self, true, session))
    }

    /// The client requests a shell.
    #[allow(unused_variables)]
    async fn shell_request(
        self,
        channel_id: ChannelId,
        mut session: Session,
    ) -> Result<(Self, Session), Self::Error> {
        debug!("shell_request");

        // mark request success
        session.request_success();

        // Spawn a new /bin/bash process
        let mut child = Command::new("/bin/bash")
[...]

PR #216, opened 2023-11-25
/bin/bash pty example

brandonros:
closes https://github.com/warp-tech/russh/issues/215

I'd expect most people to think "ssh server" as "provides shell" / "can execute shell commands"

Only problem with this one is... it accepts any password :D (and it's hardcoded to a MacOS specific branch of pty-process right now)


brandonros:
moved to https://github.com/brandonros/rustbear with @samuela's help

PR #222, opened 2023-12-18
Avert the race between sending data and sending EOF

### LINKED ISSUE #213: channel.data drops end of data
I reproduced the problem by pulling it out of my application and into a russh example: https://github.com/rukai/russh/commit/442db2bea50a76dff545b4ebd587a13a0353cbd5

I have an implementation to write to a remote file via sending data to a remote `cat > filename`
The command completes, which I can verify by seeing that the chmod 777 has succeeded.
However on some runs the end of the file is missing.
Usually this occurs within the 50 iterations in the modified example.

From my local machine I run a command like:
`cargo run --release --example client_exec_simple -- -k ~/.ssh/id_ed25519 --username rukai pi.lan`
Which connects to my raspberry pi.
I think `--release` is important.

The application I originally encountered this on is used to setup machines in AWS.

Here is an example failed run:
```
~2/Projects/Crates/russh> cargo run --release --example client_exec_simple -- -k ~/.ssh/id_ed25519 --username rukai pi.lan
    Finished release [optimized] target(s) in 0.07s
     Running `target/release/examples/client_exec_simple -k /home/rukai/.ssh/id_ed25519 --username rukai pi.lan`
[2023-11-09T03:52:12Z INFO  client_exec_simple] Connecting to pi.lan:22
[2023-11-09T03:52:12Z INFO  client_exec_simple] Key path: "/home/rukai/.ssh/id_ed25519"
[2023-11-09T03:52:13Z INFO  client_exec_simple] Connected
attempt #0 10485760 foo
attempt #1 10485760 foo
attempt #2 10485760 foo
attempt #3 10485760 foo
attempt #4 10485760 foo
attempt #5 10485760 foo
attempt #6 10485760 foo
attempt #7 10485760 foo
attempt #8 10485760 foo
attempt #9 10485760 foo
attempt #10 10485760 foo
attempt #11 10485760 foo
attempt #12 10485760 foo
attempt #13 10485760 foo
attempt #14 10485760 foo
attempt #15 10485760 foo
attempt #16 10485760 foo
attempt #17 10485760 foo
attempt #18 10485760 foo
attempt #19 10485760 foo
attempt #20 10485760 foo
attempt #21 10485760 foo
attempt #22 10436608 foo
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `10436608`,
 right: `10485760`', russh/examples/client_exec_simple.rs:48:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
   4: client_exec_simple::main::{{closure}}
   5: tokio::runtime::park::CachedParkThread::block_on
   6: tokio::runtime::context::runtime::enter_runtime
   7: tokio::runtime::runtime::Runtime::block_on
   8: client_exec_simple::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

mmirate:
This fixes #213. Also, while verifying as such, I noticed that an unrelated modification to client_exec_simple was required for correct operation, and have included it here. Namely: the receiving event-loop should not terminate early upon receiving the exit status, since SSH servers can and will send data after sending the exit status.

Also, I did something similar for the `close` operation.

Also, my initial attempt at this involved generalizing the pending_data queue so that it could contain EOF packets. That doesn't work at all, but I left in-place a refactor which should stop outgoing extended data from being transformed into regular data when placed into the pending deque.

In theory, an extra flag to restore the previous behavior might be desirable in case the previous behavior happens to be desirable; but that would be a not-strictly-necessary breaking change to the public API.

I see one downside. Consider the usecase - albeit very odd imho - where a channel has multiple owners that all constantly send a large volume of data (such that the pending deque does not empty itself as long as the sending continues) and do not cooperate among themselves as to when the channel is done being sent data. In this scenario, the status quo means that the first owner who wants an EOF will indeed cause the EOF to be sent, promptly but imprecisely. Whereas with this PR's change, the EOF might not be sent until the owners unanimously agree to cease sending data. The only way to tolerate this that I can fathom, would be to add some way to notify the user that a given channel's pending data buffer has been completely flushed - owners of singly-owned channels who have scheduled all outgoing data could await that message before sending EOF, and owners of multiply-owned channels would remain free to send EOF as soon as they please.

### DIFF
--- a/russh/examples/client_exec_simple.rs
+++ b/russh/examples/client_exec_simple.rs
@@ -105,7 +105,7 @@ impl Session {
         let mut channel = self.session.channel_open_session().await?;
         channel.exec(true, command).await?;
 
-        let mut code = 0;
+        let mut code = None;
         let mut stdout = tokio::io::stdout();
 
         loop {
@@ -121,14 +121,13 @@ impl Session {
                 }
                 // The command has returned an exit code
                 ChannelMsg::ExitStatus { exit_status } => {
-                    code = exit_status;
-                    channel.eof().await?;
-                    break;
+                    code = Some(exit_status);
+                    // cannot leave the loop immediately, there might still be more data to receive
                 }
                 _ => {}
             }
         }
-        Ok(code)
+        Ok(code.expect("program did not exit cleanly"))
     }
 
     async fn close(&mut self) -> Result<()> {
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -738,6 +738,8 @@ impl Session {
                         confirmed: true,
                         wants_reply: false,
                         pending_data: std::collections::VecDeque::new(),
+                        pending_eof: false,
+                        pending_close: false,
                     };
 
                     let confirm = || {
--- a/russh/src/lib.rs
+++ b/russh/src/lib.rs
@@ -473,6 +473,8 @@ pub(crate) struct ChannelParams {
     pub confirmed: bool,
     wants_reply: bool,
     pending_data: std::collections::VecDeque<(CryptoVec, Option<u32>, usize)>,
+    pending_eof: bool,
+    pending_close: bool,
 }
 
 impl ChannelParams {
--- a/russh/src/server/encrypted.rs
+++ b/russh/src/server/encrypted.rs
@@ -1090,6 +1090,8 @@ impl Session {
             confirmed: true,
             wants_reply: false,
             pending_data: std::collections::VecDeque::new(),
+            pending_eof: false,
+            pending_close: false,
         };
 
         let (channel, reference) = Channel::new(
--- a/russh/src/session.rs
+++ b/russh/src/session.rs
@@ -66,6 +66,23 @@ pub(crate) struct CommonSession<Config> {
     pub strict_kex: bool,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum ChannelFlushResult {
+    Incomplete { wrote: usize, },
+    Complete { wrote: usize, pending_eof: bool, pending_close: bool, }
+}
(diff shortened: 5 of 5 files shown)


Eugeny:
Do you think it would make sense in this context to simply refuse sending more data once `pending_eof` has been set? This is in line with the protocol spec and avoids the case you're describing. There'll be a breaking change since `.data()` would have to return a `Result` but that's fine IMO.

mmirate:
> Do you think it would make sense in this context to simply refuse sending more data once `pending_eof` has been set? This is in line with the protocol spec and avoids the case you're describing. There'll be a breaking change since `.data()` would have to return a `Result` but that's fine IMO.

That would be ideal if there was a way to raise that error at Channel::data (e.g. by having some way for a Channel to have an eventually-consistent view of the `pending_eof` flag of the Session to which it is connected). Since the underlying communication from the Channel to the Session is an mpsc channel, any call to Channel::data who ends up perpetrating an error at {client,server}::Session::data cannot actually receive that error and will likely have already returned to its caller by the time the error happens.

Eugeny:
What if we let ChannelParams/Channel share `pending_eof` through a `Mutex<Arc>` similarly to how `ChannelTx` instances already share their `window_size`s?

There's still one edge case left where the user calls `.eof()` and then `.data()` again while the EOF message is still queued in the mpsc channel, but this behaviour is against the spec anyway and that data could be simply discarded by the Session IMO.

PR #231, opened 2024-01-24
Add `russh::server::run_on_socket` to facilitate dropping privileges immediately after socket binding

samuela:
Securing SSH servers is not straightforward. Most state-of-the-art systems utilize privilege dropping after they bind to a socket in order to minimize their attack surface, including OpenSSH. Currently russh does not support this pattern since socket binding and the main server event loop are tied to a single function, `russh::server::run`.

This PR introduces a function `russh::server::run_on_socket` which is identical to `russh::server::run` except that it accepts a `tokio::net::TcpListener` in place of a socket address, allowing end-users to bind to a socket and perform other intermediate actions before handing over control to the main server event loop. `russh::server::run`'s behavior is left untouched, so the change is fully backwards compatible.

### DIFF
--- a/russh/src/server/mod.rs
+++ b/russh/src/server/mod.rs
@@ -630,15 +630,13 @@ pub trait Server {
     fn handle_session_error(&mut self, _error: <Self::Handler as Handler>::Error) {}
 }
 
-/// Run a server.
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

@all-contributors add @samuela for code

PR #233, opened 2024-01-26
Request to add Diffie-Hellman group 16 512

bho01:
I would like to add Diffie-Hellman group 16 512 to the list of key exchanges. Ran the test and tried on a mini client that attempted to make a request to a server that only accepts this key exchange and DHGroup18 512. It ran fine and was able to establish a connection.


Eugeny:
Thank you! 

@all-contributors add @bho01 for code

Issue #232, opened 2024-01-26
`russh::server::run` doesn't respect keepalive responses sent by the client


samuela:
I'm finding that `russh::server::run` disregards keepalive responses sent by the client and disconnects anyways. For example, I've configured the server to send keepalive packets:
```rs
  let config = russh::server::Config {
    keys: vec![...],
    keepalive_interval: Some(std::time::Duration::from_secs(1)),
    keepalive_max: 3,
    ..Default::default()
  };
```
I can confirm that these keepalive packets are received and responded to by the client:
```
❯ ssh -vvv bitbop.io
...
debug2: channel 0: request shell confirm 1
debug3: send packet: type 98
debug3: client_repledge: enter
debug2: channel_input_open_confirmation: channel 0: callback done
debug2: channel 0: open confirm rwindow 2097152 rmax 32768
┌   bitbop.io
│
◆  What type of machine would you like?
│  ● t3a.micro      2 vCPU    1G Mem                              (free!)
│  ○ g4dn.xlarge    4 vCPU   16G Mem    1x T4   GPU   16G GPU Mem
│  ○ p3.2xlarge     8 vCPU   61G Mem    1x V100 GPU   16G GPU Mem
│  ○ p4d.24xlarge  96 vCPU  1.2T Mem    8x A100 GPU  320G GPU Mem
│  ○ p5.48xlarge  192 vCPU    2T Mem    8x H100 GPU  640G GPU Mem
└
debug3: receive packet: type 80
debug1: client_input_global_request: rtype keepalive@openssh.com want_reply 1
debug3: send packet: type 82
debug3: receive packet: type 80
debug1: client_input_global_request: rtype keepalive@openssh.com want_reply 1
debug3: send packet: type 82
debug3: receive packet: type 80
debug1: client_input_global_request: rtype keepalive@openssh.com want_reply 1
debug3: send packet: type 82
debug3: receive packet: type 80
debug1: client_input_global_request: rtype keepalive@openssh.com want_reply 1
debug3: send packet: type 82
debug3: send packet: type 1
debug1: channel 0: free: client-session, nchannels 1
debug3: channel 0: status: The following connections are open:
  #0 client-session (t4 [session] r2 i0/0 o0/0 e[write]/0 fd 6/7/8 sock -1 cc -1 io 0x01/0x00)
```
But the connection is still getting killed... I'm interpreting `debug3: send packet: type 82` as a keepalive response sent by the client. 

[...]

PR #243, opened 2024-02-04
Fix https://github.com/warp-tech/russh/issues/232

samuela:
Upon debugging, I discovered that `alive_timeouts` is not currently ever reset to 0 anywhere. The diff in `server/session.rs` remedies this, reseting `alive_timeouts` to 0 any time we receive data from the client.

Unfortunately, there was additionally the issue that `self.common.received_data` was not in fact a reliable indicator of whether or not we had actually received data from the client. This turned out to be due to some logic in `server/encrypted.rs` that overrode the assignment [here](https://github.com/warp-tech/russh/blob/c7f6c5b867fd063c383a7e07acd311ffce0d1c80/russh/src/server/session.rs#L403) if the message was not one of the expected format. I have removed this logic in `server/encrypted.rs`.

`self.common.received_data` is not documented or defined AFAIK. Looking through the code, I concluded that `server/encrypted.rs` was in the wrong and should be modified, but some scrutiny on this decision may be warranted. In any case, this fixes https://github.com/warp-tech/russh/issues/232.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -500,6 +500,14 @@ impl Session {
                 .map_err(crate::Error::from)?;
             self.common.write_buffer.buffer.clear();
 
+            if self.common.received_data {
+                // Reset the number of failed keepalive attempts. We don't
+                // bother detecting keepalive response messages specifically
+                // (OpenSSH_9.6p1 responds with REQUEST_FAILURE aka 82). Instead
+                // we assume that the client is still alive if we receive any
+                // data from it.
+                self.common.alive_timeouts = 0;
+            }
             if self.common.received_data || sent_keepalive {
                 if let (futures::future::Either::Right(ref mut sleep), Some(d)) = (
                     keepalive_timer.as_mut().as_pin_mut(),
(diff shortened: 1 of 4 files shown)


Eugeny:
@mmirate could you please review this as well? Thanks.

russh/src/server/session.rs
```diff
@@ -500,6 +500,14 @@ impl Session {
...
             self.common.write_buffer.buffer.clear();
 
+            if self.common.received_data {
+                // Reset the number of failed keepalive attempts. We don't
+                // bother detecting keepalive response messages specifically
+                // (OpenSSH_9.6p1 responds with REQUEST_FAILURE aka 82). Instead
+                // we assume that the client is still alive if we receive any
+                // data from it.
+                self.common.alive_timeouts = 0;
+            }
```
mmirate, line 510: Good catch. This ought to be present on the client as well.

mmirate (changes requested):
This is the correct way forward. It just needs one specific additional piece of follow-through.

russh/src/server/session.rs
samuela: Roger that, similar changes applied to the client code in https://github.com/warp-tech/russh/pull/243/commits/ab9d6c74c70415c5c7befb2c16857fe301f42fe1

PR #191, opened 2023-09-26
Refactor the Handler traits based on feedbacks on #188

lowlevl:
This is a PR based on the feedback received in #188, where I proposed some improvement on the overall API of the `russh` crate.

PR #247, opened 2024-02-07
Refactor the Handler traits

ricott1:
Follow up to https://github.com/warp-tech/russh/pull/191.

- Remove the owned pattern with tuple return for the Handlers traits in favour of &mut references.

### DIFF
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -77,7 +77,6 @@ mod session;
 /// It is in charge of multiplexing and keeping track of various channels
 /// that may get opened and closed during the lifetime of an SSH session and
 /// allows sending messages to the server.
-
 pub struct Session {
     common: CommonSession<Arc<Config>>,
     receiver: Receiver<Msg>,
@@ -1110,7 +1109,7 @@ thread_local! {
 
 impl KexDhDone {
     async fn server_key_check<H: Handler>(
-        &mut self,
+        mut self,
         rekey: bool,
         handler: &mut H,
         buf: &[u8],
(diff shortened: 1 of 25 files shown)


ricott1:
I was also thinking about moving `run` and `run_on_socket` to the `Server` trait, I will do that in a separate commit, what do you think?

russh/src/client/mod.rs
```diff
@@ -77,6 +77,7 @@ mod session;
 /// It is in charge of multiplexing and keeping track of various channels
 /// that may get opened and closed during the lifetime of an SSH session and
 /// allows sending messages to the server.
+
```
lowlevl, line 80: Not sure if this newline is intentional ?

russh/src/client/mod.rs
```diff
@@ -1111,11 +1110,11 @@ thread_local! {
 
 impl KexDhDone {
     async fn server_key_check<H: Handler>(
-        mut self,
+        &mut self,
```
lowlevl, line 1113: I'm not sure to understand why you made this change, since it implies a bunch of things to be cloned, while it seems the `KexDhDone` seems to be dropped after this call everytime anyways, I'd suggest still taking `self` and reverting the clone changes you did in `KexAlgorithm` and `KexDhDone::compute_keys`.

russh/src/client/mod.rs
ricott1: nope, removed

russh/src/client/mod.rs
ricott1: You're right, it's not necessary, I reverted it

lowlevl:
I guess it looks good to me, but as I said, I'm not the one to decide :)

Eugeny:
great job @ricott1 and thanks for reviewing @lowlevl !

Eugeny:
@all-contributors add @ricott1 for code

PR #249, opened 2024-02-11
Add Ratatui apps over SSH to examples

ricott1:
We could add some minimal examples on how to serve a [ratatui](https://github.com/ratatui-org/ratatui) app over SSH. After running the example, connect the clients by running `ssh localhost -p 2222`.

The `ratatui_app.rs` example serves a separate app instance to each client, which can be checked by connecting 2 clients and trying to reset the app counter by pressing `c`. Each app counter should update independently.

The `ratatui_shared_app.rs` example serves the same app instance to each client, which can be checked by connecting 2 clients and trying to reset the app counter by pressing `c`. Each client should see the app counter update at the same time. This example can be used to build something along the lines of [sshtron](https://github.com/zachlatta/sshtron).

I also update the toolchain file to fix a build error, let me know if you prefer to have that committed to a separate PR.

### DIFF
--- a/russh/Cargo.toml
+++ b/russh/Cargo.toml
@@ -73,6 +73,7 @@ rand = "0.8.5"
 shell-escape = "0.1"
 tokio-fd = "0.3"
 termion = "2"
+ratatui = "0.26.0"
 
 [package.metadata.docs.rs]
 features = ["openssl"]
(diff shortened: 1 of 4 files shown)


ricott1:
The unsuccessful check is an unrelated clippy issue

Eugeny:
Thank you, merged!

PR #255, opened 2024-02-26
Additional Disconnect Handling

amtelekom:
- Allow Server Handle to trigger Disconnect
- Allow Client Handler to read Disconnect reason

### DIFF
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -36,6 +36,7 @@
 
 use std::cell::RefCell;
 use std::collections::{HashMap, VecDeque};
+use std::convert::TryInto;
 use std::num::Wrapping;
 use std::pin::Pin;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


Eugeny:
I think the naming of `Handler::disconnect` is going to be confusing for the users - just today there was a user wondering how to react to session closure due to an error. Most will probably assume that it's going to be called when the session is closed for any reason - what do you think?

Eugeny:
Otherwise LGTM

amtelekom:
I see the problem. I would still use the word `disconnect`, since that's what RFC 4253 calls it.

I can see 2 Variants:
- prefix it with `received_` to hopefully make more clear we're just talking about a normal session end ([22373e5](https://github.com/warp-tech/russh/pull/255/commits/22373e5dd4a9f3e0bc8477b72f2e987a1e5d87be))
  This now opens up the possibility to add a `disconnect` or `connection_closed` callback that is called if the main `tokio::select!` loop is left for any reason, or users can parse the result of the join. (

- Pipe all kinds of disconnects through the disconnected handler ([2e39628](https://github.com/warp-tech/russh/pull/255/commits/2e396282d16e36fdaac948667a9f073c750314da)). Since the error isn't `Clone` and a borrow to it wouldn't be `Sync`, we have to move the Error through the callback in this case though.

Eugeny:
The 2nd variant is much clearer to the user and is universal, I like it! :+1: :+1: 

Now just to remove this comment https://github.com/warp-tech/russh/pull/255/files#diff-a54f5b4b2c7b6955dd496e3f9b7bcc52dac6226eb3450c4785bd7957e47f0c1dR1694 and the PR is good to go

amtelekom:
Fixed & squashed

Eugeny:
Thanks - appreciate you working on these things :v:

PR #263, opened 2024-03-27
Fix compilation error when using without the flate2 feature

citorva:
When compiling without the default features, the compilation of russh fail with message concerning unknown Error type in compression.rs and unknown flate2 crate in lib.rs (when defining the error type)

This PR specify the same Error type as the flate2-featured Compression/Decompression function and disable error cases specific to flate2.


Eugeny:
Thank you!

Eugeny:
@all-contributors add @citorva for code

Issue #271, opened 2024-04-23
`write_key_v1` writes `ed25519` key incorrectly


robertabcd:
In the effort of replacing `yasna` with `der` crate (which conveniently `pkcs1`, `sec1`, `ed25519` integrate with to encode and decode DER), these 2 bugs were discovered.

1. `publicKey` should be written with an IMPLICIT tag

Per ASN.1 module in [RFC 5958 Appendix A](https://datatracker.ietf.org/doc/html/rfc5958#appendix-A), devil in the beginning:

```
DEFINITIONS IMPLICIT TAGS ::=

   BEGIN

...

OneAsymmetricKey ::= SEQUENCE {
     version                   Version,
     privateKeyAlgorithm       PrivateKeyAlgorithmIdentifier,
     privateKey                PrivateKey,
     attributes            [0] Attributes OPTIONAL,
     ...,
     [[2: publicKey        [1] PublicKey OPTIONAL ]],
     ...
   }
```

2. Only the secret part is written in `privateKey`.

An example in [RFC 8410 Section 10.3](https://datatracker.ietf.org/doc/html/rfc8410#section-10.3)

The `ed25519` crate implements correctly and will fail to decode keys written by us.

robertabcd:
Note that I can't make `ssh-keygen` or `openssl genpkey` write `ed25519` keys in the `OneAsymmetricKey` with `publicKey` specified, so I'm not sure how well the inter-op is.

PR #272, opened 2024-04-26
Fix ed25519 key format in pkcs8

robertabcd:
* Add support for decoding ed25519 keys in PrivateKeyInfo format
* Fix ed25519 encoding and decoding in OneAsymmetricKey format

Closes: https://github.com/warp-tech/russh/issues/271


Eugeny:
Thank you!

Issue #225, opened 2024-01-10
Consider replacing OpenSSL with pure-rust alternative


cipriancraciun:
At the moment, from what I gather, the only requirement for the OpenSSL dependency is to support the RSA algorithms, mainly in the `russh-keys` crate, and a bit in `russh` itself.

However, the main issue with OpenSSL is cross-compilation, as for example in my own project the most cross compilation issues I had was related with the OpenSSL library.

Thus, it would be nice to either replace OpenSSL with a pure Rust implementation, or have it as an alternative.  For example RusTLS depends on the <https://crates.io/crates/rsa> crate which is written in pure Rust.

----

Now, given that for Ed25519 keys this project uses Rust native implementations, my assumption is that OpenSSL was an initial dependency, and thus it remains to this day until the code in question is rewritten for an alternative.

cipriancraciun:
Looking at the existing pull-requests, it's somewhat the opposite of #52, that wants to rely only on OpenSSL cryptography.

I think having both options would solve opposite requirements (those wanting a pure-Rust implementation, and those wanting OpenSSL).

However, implementing both, I think would complicate the code a bit, as it would litter the code with `#[cfg(...)]` items...

Perhaps, a sub-crate should be created, say `russh-crypto` that provides a wrapper for whatever backend the user enables via features (say, OpenSSL, or pure-Rust), and having the rest of the `russh*` crates not touch cryptographic libraries directly.

Eugeny:
Yep I'd like to integrate rust-crypto's RSA impl as well and have it as the default set. Cipher modules are actually fairly pluggable in the way how russh is structured so it'd just be a single `#[cfg]` per module. Just haven't had time to work on it.

cipriancraciun:
@Eugeny although I don't promise anything (my "hobby budget" is already in the red), if I have the time I'll try to see if maybe I can provide a pull request.

One question though:  should I remove the need for OpenSSL, or should pure-Rust RSA implementation be an alternative to OpenSSL?  (If there isn't a good motive to keep OpenSSL in, I would suggest removing it, as it would keep the code simpler.)

Eugeny:
That would be amazing! I want to keep openSSL in, as currently russh is used in VSCode where Microsoft maintains their own fork that uses OpenSSL RSA and also adds OpenSSL AES on top of russh.

[...]

PR #273, opened 2024-04-27
Add pure-rust RSA implementation

robertabcd:
* SSH encoding of RSA keys is moved into `protocol` module.
* Decouple the SSH encoding into traits.
* When `openssl` feature is not enabled, the pure-rust RSA impl is used.

Alternative implementation for https://github.com/warp-tech/russh/issues/225


Eugeny:
Fantastic 💪

PR #288, opened 2024-05-25
cryptovec: Fix a segmentation fault

elegaanz:
The `CryptoVec::resize` implementation was running into a segmentation fault in some case. If the capacity of the vector was more than 0, and that the new allocation failed, the call to `std::ptr::copy_non_overlapping` would have a null pointer as destination. This was very easy to trigger by a malicious peer, they just had to send a packet with an announced size large enough for the allocation to fail. The code now correctly panics, which would only end the current thread and not crash the whole application without giving it a chance to continue running.


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @elegaanz for code

PR #290, opened 2024-05-26
Allow ssh-rsa keys to be used for rsa-sha2-* auth

elegaanz:
This worked before [this change](https://github.com/warp-tech/russh/commit/194430b2326a7681a25f5a5f4d5b72749c128fe8#diff-81b0128d04a2d9a37bdc6931dec1cb426a8ff7c6ad4ef10e1555ca3e29f08380L159), and is permitted by the specification, as can be seen in [the example in section 3.2 of RFC8332](https://www.rfc-editor.org/rfc/rfc8332#section-3.2).


Eugeny:
LGTM, thank you for the improvements!

PR #294, opened 2024-06-01
Fix hardcoded public key auth negotiation

tomknig:
Authentication using RSA public keys was incorrectly hardcoded to use `ssh-rsa`, i.e., SHA1. This led to public keys set up to use `rsa-sha2-256` or `rsa-sha2-512` falling back to `ssh-rsa` and being rejected by modern SSH servers that are configured not to support SHA1 by default.

The solution derives the hash name from the public key instead of hardcoding it.

### DIFF
--- a/russh/src/negotiation.rs
+++ b/russh/src/negotiation.rs
@@ -138,13 +138,12 @@ impl Named for () {
 }
 
 use russh_keys::key::ED25519;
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! 

@all-contributors please add @tomknig for code

tomknig:
Thanks @Eugeny!

PR #306, opened 2024-06-28
Add support for glob pattern matching in Host directives

packetsource:
- update parse logic to perform host glob matching (using globset)
- generalise token expansion so it can be used to qualify hosts as well as localising ProxyCommand
- add proxyjump to the config structure


Eugeny:
Thank you! 

@all-contributors please add @packetsource for code

Issue #311, opened 2024-07-06
streamlocal-forward support


kanpov:
Remote UDS => local UDS

kanpov:
I'll try to contribute

kanpov:
```
Similar to tcpip-forward, streamlocal-forward is sent by the client
to request remote forwarding of a Unix domain socket.

	byte		SSH2_MSG_GLOBAL_REQUEST
	string		"streamlocal-forward@openssh.com"
	boolean		TRUE
	string		socket path

Similar to cancel-tcpip-forward, cancel-streamlocal-forward is sent
by the client cancel the forwarding of a Unix domain socket.

	byte		SSH2_MSG_GLOBAL_REQUEST
	string		"cancel-streamlocal-forward@openssh.com"
	boolean		FALSE
	string		socket path
```

kanpov:
@Eugeny Tried to do this in #312

PR #312, opened 2024-07-06
Implement streamlocal-forward for remote => local UDS forwarding

kanpov:
I left a `// NEED HELP` comment on places where I didn't fully figure out what to do, so I'd really appreciate it if some maintainers helped me out in those places.

### DIFF
--- a/russh/src/client/session.rs
+++ b/russh/src/client/session.rs
@@ -319,6 +319,48 @@ impl Session {
         }
     }
 
+    pub fn streamlocal_forward(
+        &mut self,
+        reply_channel: Option<oneshot::Sender<Option<String>>>,
+        socket_path: &str,
+    ) {
+        if let Some(ref mut enc) = self.common.encrypted {
+            let want_reply = reply_channel.is_some();
+            if let Some(reply_channel) = reply_channel {
+                self.open_global_requests.push_back(
+                    crate::session::GlobalRequestResponse::StreamLocalForward(reply_channel),
+                );
+            }
+            push_packet!(enc.write, {
+                enc.write.push(msg::GLOBAL_REQUEST);
+                enc.write.extend_ssh_string(b"streamlocal-forward");
+                enc.write.push(want_reply as u8);
+                enc.write.extend_ssh_string(socket_path.as_bytes());
+            });
+        }
+    }
+
+    pub fn cancel_streamlocal_forward(
+        &mut self,
+        reply_channel: Option<oneshot::Sender<bool>>,
+        socket_path: &str,
+    ) {
+        if let Some(ref mut enc) = self.common.encrypted {
+            let want_reply = reply_channel.is_some();
+            if let Some(reply_channel) = reply_channel {
+                self.open_global_requests.push_back(
+                    crate::session::GlobalRequestResponse::CancelStreamLocalForward(reply_channel),
+                );
+            }
+            push_packet!(enc.write, {
+                enc.write.push(msg::GLOBAL_REQUEST);
+                enc.write.extend_ssh_string(b"cancel-streamlocal-forward");
+                enc.write.push(want_reply as u8);
+                enc.write.extend_ssh_string(socket_path.as_bytes());
+            });
+        }
+    }
+
     pub fn send_keepalive(&mut self, want_reply: bool) {
         self.open_global_requests
             .push_back(crate::session::GlobalRequestResponse::Keepalive);
--- a/russh/src/parsing.rs
+++ b/russh/src/parsing.rs
@@ -93,6 +93,7 @@ pub enum ChannelType {
     },
     DirectTcpip(TcpChannelInfo),
     ForwardedTcpIp(TcpChannelInfo),
+    ForwardedStreamLocal(StreamLocalChannelInfo),
     AgentForward,
     Unknown {
         typ: Vec<u8>,
@@ -105,6 +106,12 @@ pub struct TcpChannelInfo {
     pub port_to_connect: u32,
     pub originator_address: String,
     pub originator_port: u32,
+}
+
+#[derive(Debug)]
+pub struct StreamLocalChannelInfo {
+    pub client_socket_path: String,
+    pub server_socket_path: String,
 }
 
 impl TcpChannelInfo {
--- a/russh/src/server/encrypted.rs
+++ b/russh/src/server/encrypted.rs
@@ -1026,6 +1026,14 @@ impl Session {
                         }
                         Ok(())
                     }
+                    b"streamlocal-forward" => {
+                        // NEED HELP
+                        Ok(())
+                    }
+                    b"cancel-streamlocal-forward" => {
+                        // NEED HELP
+                        Ok(())
+                    }
                     _ => {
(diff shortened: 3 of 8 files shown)


kanpov:
I seem to have implemented most if not all of the needed behavior (but not tests yet), but this needs thorough review by people who know the library better than me.

kanpov:
@Eugeny would really appreciate it if you took a look

kanpov:
Fixes #311

russh/src/server/session.rs
```diff
@@ -919,6 +986,18 @@ impl Session {
...
 
+    pub fn channel_open_forwarded_streamlocal(
+        &mut self,
+        server_socket_path: &str,
+        client_socket_path: &str,
+    ) -> Result<ChannelId, Error> {
+        self.channel_open_generic(b"forwarded-streamlocal", |write| {
+            // NEED HELP: is this correct?
+            write.extend_ssh_string(server_socket_path.as_bytes());
+            write.extend_ssh_string(client_socket_path.as_bytes());
```
Eugeny, line 997: Looks like this field is just reserved: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL#L344

russh/src/parsing.rs
```diff
@@ -107,6 +111,29 @@ pub struct TcpChannelInfo {
...
+    pub server_socket_path: String,
+}
+
+impl StreamLocalChannelInfo {
+    fn new(r: &mut Position) -> Result<Self, crate::Error> {
+        // NEED HELP: is this completely correct?
+        let client_socket_path = std::str::from_utf8(r.read_string().map_err(crate::Error::from)?)
+            .map_err(crate::Error::from)?
+            .to_owned();
+        let server_socket_path = std::str::from_utf8(r.read_string().map_err(crate::Error::from)?)
```
Eugeny, line 126: Same here

russh/src/client/session.rs
```diff
@@ -319,6 +319,56 @@ impl Session {
...
+        if let Some(ref mut enc) = self.common.encrypted {
+            let want_reply = reply_channel.is_some();
+            if let Some(reply_channel) = reply_channel {
+                self.open_global_requests.push_back(
+                    crate::session::GlobalRequestResponse::StreamLocalForward(reply_channel),
+                );
+            }
+            push_packet!(enc.write, {
+                enc.write.push(msg::GLOBAL_REQUEST);
+                enc.write.extend_ssh_string(b"streamlocal-forward");
```
Eugeny, line 340: Should be namespaced to @openssh.com: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL#L355

Eugeny (changes requested):

russh/src/client/session.rs
kanpov: Done, streamlocal-forward => streamlocal-forward@openssh.com and the same for cancel-streamlocal-forward

russh/src/server/session.rs
kanpov: Pretty sure "socket path" is the client socket path

kanpov: So this should be correct:
```rust
            write.extend_ssh_string(client_socket_path.as_bytes());
            write.extend_ssh_string(b"");
```

kanpov: Probably resolved? I'm kinda confused though since tcp/ip forwarding has both client and server (host/port) but streamlocal forwarding only has client?

russh/src/parsing.rs
kanpov: Then how to determine server_socket_path? Or should it not even be known? Once again, confused since tcp/ip has both the client's and the server's, but apparently streamlocal only has one?

kanpov:
I'm pretty sure I handled the reserved field correctly, but not completely sure

russh/src/server/encrypted.rs
```diff
@@ -1026,6 +1026,41 @@ impl Session {
                         }
                         Ok(())
                     }
+                    b"streamlocal-forward@openssh.com" => {
+                        let server_socket_path =
+                            std::str::from_utf8(r.read_string().map_err(crate::Error::from)?)
+                                .map_err(crate::Error::from)?;
+                        debug!("handler.streamlocal_forward {:?}", server_socket_path);
+                        let mut client_socket_path = String::from(server_socket_path).clone();
```
Eugeny, line 1034: client_socket_path is a duplicate of server_socket_path here

Eugeny:
I think you can go ahead and try it out :) looks mostly fine so far

russh/src/server/encrypted.rs
kanpov: Isn't the point of this so that the server can decide which client_socket_path to return? (based on a handler method accepting server_socket_path)

kanpov:
> I think you can go ahead and try it out :) looks mostly fine so far

I'll try by including the crate by path, hope it works

kanpov:
Update: it doesn't work :( RequestDenied, gonna try to investigate

Eugeny:
`streamlocal-forward` is `-R` :)

kanpov:
Well `-L` also works for some reason

kanpov:
@Eugeny Makes no sense, where's the issue here?

Impl:
```rust
            push_packet!(enc.write, {
                enc.write.push(msg::GLOBAL_REQUEST);
                enc.write
                    .extend_ssh_string(b"streamlocal-forward@openssh.com");
                enc.write.push(want_reply as u8);
                enc.write.extend_ssh_string(socket_path.as_bytes());
            });
```

Spec:
```
	byte		SSH2_MSG_GLOBAL_REQUEST
	string		"streamlocal-forward@openssh.com"
	boolean		TRUE
	string		socket path
```

kanpov:
@Eugeny While testing this, I also tested tcp/ip forwarding with `handle.tcpip_forward` and even that didn't work, even though a port was returned that port didn't have any connectivity.

kanpov:
@Eugeny Considering forwarding is broken, I'm pretty sure making it work is not as simple as just sending global messages to enable/cancel forwarding over SSH.

My brain is too smooth to understand openssh's code, but I suppose what they're doing is not just making these global requests (which russh is capable of doing correctly, with some caveats), but also actually **binding** to these sockets itself and proxying all traffic over.

Unfortunately, there is little reference aside from openssh C code (which I'll still try to dig into sometime) about how exactly this is done, since openssh is one of the only implementations of ssh that actually support forwarding.

kanpov:
@Eugeny Have you tested whether this actually works after your fixes? Sorry for abandoning the PR

russh/src/server/encrypted.rs
Eugeny: No, there is no second path anywhere in the protocol

Eugeny:
Fixed a few issues and now it works:
* Removed duplicate implementation for non-standard client-to-server forwarding channels
* Added the missing @openssh.com in channel types and responses
* The server doesn't return a socket path in the response

Eugeny:
No worries! It works with simple client & server examples - but I just tested it briefly

PR #320, opened 2024-07-31
Consolidate dirs deps

rukai:
It seems like the project ended up with two different dependencies providing access to the home directory.
This PR consolidates these dependencies into a single dependency.

I've done so by using the `home` crate since on linux it has 0 dependencies and is well supported due to being a part of the cargo repo.
But if you'd prefer I could swap this PR to use `dirs-next`.

I would like to avoid `dirs` since it is pulling in strange dependencies https://github.com/dirs-dev/dirs-sys-rs/pull/26

### DIFF
--- a/russh-config/Cargo.toml
+++ b/russh-config/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.7.1"
 rust-version = "1.65"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


Eugeny:
Thanks!

PR #328, opened 2024-08-21
Remove leftover extraneous debugging in host globbing function

packetsource:
Apologies. I accidentally left a stray dbg!() in when checking the glob expansion logic of hosts in the russh-config SSH config parsing code. This PR simply removes it.


Eugeny:
No problem, thanks!

PR #331, opened 2024-08-24
fix: process multiple host entries

CertainLach:
PR https://github.com/Eugeny/russh/pull/306 broke russh config parsing for me.

My ssh config looks roughly like this:

```
Host *
    Common settings
Host wanted
   Host-specific settings
```

Before https://github.com/Eugeny/russh/pull/306 common settings were ignored, but after - common settings were applied, yet config parser stops right on next `Host` section start, so settings I need were ignored instead.

This PR brings behavior closer to what openssh itself does - for requested host, it applies all the settings with maching glob, not only the first one.

### DIFF
--- a/russh-config/src/lib.rs
+++ b/russh-config/src/lib.rs
@@ -113,15 +113,20 @@ pub enum AddKeysToAgent {
 }
 
 pub fn parse(file: &str, host: &str) -> Result<Config, Error> {
-    let mut config: Option<Config> = None;
+    let mut config = Config::default(host);
+    let mut matches_current = false;
     for line in file.lines() {
         let tokens = line.trim().splitn(2, ' ').collect::<Vec<&str>>();
         if tokens.len() == 2 {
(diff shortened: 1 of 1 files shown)


russh-config/src/lib.rs
```diff
@@ -166,22 +171,10 @@ pub fn parse(file: &str, host: &str) -> Result<Config, Error> {
...
-                config = Some(c)
             }
         }
     }
-    if let Some(config) = config {
-        Ok(config)
-    } else {
-        Err(Error::HostNotFound)
-    }
+    Ok(config)
```
CertainLach, line 177: With glob directives, it makes little sense to return HostNotFound error, I may return this error, but I think most of the time users just want to parse all relevant host settings, and will `unwrap_or_default` anyways after config parse call.

Eugeny: Makes sense

Eugeny:
Thank you!

Eugeny:
@all-contributors please add @CertainLach for code

PR #333, opened 2024-09-03
Add minimal russh ssh-agent implementation

quexten:
This PR adds a minimal ssh agent implementation, removing all parts of russh that we do not need for bitwarden-desktop.


quexten:
Apologies to anyone confused, I had my branching messed up, this was meant for a downstream fork.

PR #339, opened 2024-09-11
Improve echoserver example, bump Rust toolchain to 1.81

JuliDi:
While trying out the echoserver, I've added some error handling and made the session closable with ctrl+c, which might often be what you want. So I thought upstreaming these changes might be a good thing.

While on it, I took the liberty of bumping the rust toolchain to 1.81, which does not seem to break anything.

### DIFF
--- a/russh-keys/src/lib.rs
+++ b/russh-keys/src/lib.rs
@@ -960,7 +960,7 @@ ocyR
     fn test_decode_encode_symmetry(key: &str) {
         let original_key_bytes = data_encoding::BASE64_MIME
             .decode(
-                &key.lines()
+                key.lines()
                     .filter(|line| !line.starts_with("-----"))
                     .collect::<Vec<&str>>()
                     .join("")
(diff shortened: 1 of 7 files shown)


JuliDi:
There is still a clippy lint regarding unused struct fields:

```
warning: fields `key` and `algo` are never read
   --> russh/src/auth.rs:140:9
    |
139 |     PublicKey {
    |     --------- fields in this variant
140 |         key: CryptoVec,
    |         ^^^
141 |         algo: CryptoVec,
    |         ^^^^
    |
    = note: `CurrentRequest` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `#[warn(dead_code)]` on by default
```

Not sure what that struct is used for and how to fix it. Maybe allow dead_code in there? Or prefix with an underscore?

Regarding rustfmt: should we use a nightly toolchain to run rustfmt? There are still some warnings about nightly-only features.

Eugeny:
Thanks! Yes, the nightly `rustfmt` is still needed for `imports_granularity` etc.

Eugeny:
@all-contributors add @JuliDi for code

PR #346, opened 2024-09-19
Minor accuracy improvements to server documentation

belak:
As far as I can tell, the `RunningSession` returned by `run_stream` is only a handle to a running task, and will not kill the task when dropped as the documentation implies.


Eugeny:
Thanks! The behaviour was changed but I forgot to update the doc :+1:

@all-contributors add @belak for docs

PR #350, opened 2024-09-19
feat(wasm): Add util creates for runtime,future and time primitives for wasm

irvingoujAtDevolution:
This is the second pull request for supporting WASM.

Introducing russh-util crates to encapsulate platform difference.

Prepare for replacing tokio-util, tokio runtime and std::time::Instant::now.

### DIFF
--- a/russh-util/src/future.rs
+++ b/russh-util/src/future.rs
@@ -1,0 +1,157 @@
+use std::alloc::Layout;
+use std::fmt;
+use std::future::{self, Future};
+use std::mem::{self, ManuallyDrop};
+use std::pin::Pin;
+use std::ptr;
+use std::task::{Context, Poll};
+
+/// A reusable `Pin<Box<dyn Future<Output = T> + Send + 'a>>`.
+///
+/// This type lets you replace the future stored in the box without
+/// reallocating when the size and alignment permits this.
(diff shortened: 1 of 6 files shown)


russh-util/src/future.rs
```diff
@@ -0,0 +1,158 @@
...
+use std::mem::{self, ManuallyDrop};
+use std::pin::Pin;
+use std::ptr;
+use std::task::{Context, Poll};
+
+/// A reusable `Pin<Box<dyn Future<Output = T> + Send + 'a>>`.
+///
+/// This type lets you replace the future stored in the box without
+/// reallocating when the size and alignment permits this.
+pub struct ReusableBoxFuture<'a, T> {
```
Eugeny, line 14: Where is this going to be used?

russh-util/src/runtime.rs
```diff
@@ -0,0 +1,16 @@
+use std::future::Future;
+
+pub fn spawn<F>(future: F)
```
Eugeny, line 3: This needs to be generic over `Output` as well, as it's used in various places. I prefer this PR to include the migration of `russh` and `russh-keys` over to the generic runtime before merging.

Eugeny (changes requested):

PR #351, opened 2024-09-20
feat(wasm-support): add wasm support 

irvingoujAtDevolution:
Hi @Eugeny,

As mentioned in the previous pull request, it seems like you would prefer a complete PR for the entire integration. Hence, I have merged the previous two pull requests into one, containing the full integration. Please take a look!

### Key Changes:
1. Introduced the `russh-util` crate, and sealed the runtime with conditional compilation: **Tokio** for non-WASM, and **wasm-bindgen** for WASM.
2. Sealed the time functionality in `russh-util`: using **std::time** for non-WASM, and **chrono** for WASM.
3. Updated `cryptovec`, encapsulating the differences for `memset`, `memcpy`, `mlock`, and `munlock` into separate files, gated behind feature flags.
4. Conditionally excluded the `server` module from compiling for WASM. If anyone needs it in the future, they can follow a similar approach to the one I used.
5. Restructured the `known_hosts` module and gated it behind a compilation flag, as the WASM environment typically lacks a file system.
6. Made various minor changes to the project structure to improve maintainability.
7. Added `server`-specific `cfg` and `cfg_attr` attributes to handle platform differences in usage.

Let me know if any further modifications are needed!


russh-util/src/runtime.rs
```diff
@@ -0,0 +1,110 @@
+#[cfg(not(target_arch = "wasm32"))]
+pub use native::*;
+#[cfg(target_arch = "wasm32")]
+pub use wasm::*;
+
+#[derive(Debug)]
+pub struct JoinError {
+    #[cfg(not(target_arch = "wasm32"))]
+    inner: tokio::task::JoinError,
```
irvingoujAtDevolution, line 9: I see that the current approach with `JoinHandle` seems to align with your preference. However, I’d like to propose a discussion on possibly using a one-shot channel as a unified simulation for the handle across targets. This would significantly reduce the code complexity and maintain a consistent type between WASM and non-WASM environments. By doing so, we could simplify the overall implementation while still achieving the desired functionality.

Eugeny:
Thank you for your work! I've added some changes:

* Simplified `spawn` to always use a channel as you suggested
* Removed panic on the `JoinHandle` channel (tokio tasks do not panic if you do not await them)
* Un-gated `russh_keys::agent`
* Unified the duplicate write_kex impls
* Removed `russh::Error::TokioJoin`
* Added a CI build

If all looks good to you, this PR is good to go

irvingoujAtDevolution:
@Eugeny Yes, LGTM! Thank you very much

Eugeny:
@all-contributors add @irvingoujAtDevolution for code

PR #352, opened 2024-09-21
Reject unsupported key types instead of failing

nbdd0121:
Currently russh doesn't support ED25519/ECDSA-SK keys, but OpenSSH will attempt to use them anyway. Key parse will then fail, killing the session.

This should be treated as-if auth_publickey_offered rejects the key.


Eugeny:
LGTM, thank you!

@all-contributors add @nbdd0121 for code

PR #357, opened 2024-09-27
Update sftp examples

AspectUnk:
The current examples are a bit outdated and have some issues like https://github.com/AspectUnk/russh-sftp/issues/52#issuecomment-2380209940


Eugeny:
Thank you!

PR #360, opened 2024-10-16
Add ed25519 to ALL_KEY_TYPES

Tehforsch:
Hi - thanks for the crate!

I was trying to make my ssh client only accept `ed25519` keys. To do so, I wanted to do something like 
```rust
Preferred { 
    key: Cow::from([&key::Name::try_from("ssh-ed25519")]),
    ...
}
```
However, the `try_from` will result in `Err(())`, because `ED25519` is not contained in the `ALL_KEY_TYPES` array which is used for the lookup of all possible names.

This PR naively adds `ED25519` to `ALL_KEY_TYPES` it in there. However, there might be a good reason why this wasn't there to begin with, in which case this can of course be closed.

### DIFF
--- a/russh-keys/src/key.rs
+++ b/russh-keys/src/key.rs
@@ -54,6 +54,7 @@ pub const SSH_RSA: Name = Name("ssh-rsa");
 
 pub static ALL_KEY_TYPES: &[&Name] = &[
     &NONE,
+    &ED25519,
     &SSH_RSA,
     &RSA_SHA2_256,
     &RSA_SHA2_512,


Eugeny:
Thank you! There's no reason why it wasn't included. 

If your algo selection isn't actually dynamic, you can you can add [ED25519](https://docs.rs/russh-keys/latest/russh_keys/key/constant.ED25519.html) to the array directly instead of `try_from` so that you can keep using the stable version

@all-contributors add @Tehforsch for code

Tehforsch:
> If your algo selection isn't actually dynamic, you can you can add [ED25519](https://docs.rs/russh-keys/latest/russh_keys/key/constant.ED25519.html) to the array directly instead of try_from so that you can keep using the stable version

Thanks for the super quick reply! In my case it was actually dynamic, so I had to do the `try_from`.

PR #365, opened 2024-10-21
refactor: make `clone_public_key` infallible

nicmr:
Hi, I was wondering why `KeyPair::clone_public_key` was fallible.
Turns out only the `RsaPrivate` -> `RsaPublic` conversion used `TryFrom` internally - a `TryFrom` implementation that never failed and just returned `Ok`, so it could just be `From`.
I've switched out `TryFrom` for `From` and switched the Result type for `clone_public_key` from `Result<T>` to `T`.

I've updated the tests and doctests touching `clone_public_key` to use the new signature so they all compile and pass again. 

The only argument I see right now not to change the signature would be for forward compatibility for key types where there actually is a sensible error scenario for extracting the public key, which I find hard to imagine. Please let me know if I'm being ignorant here :thinking: 

Best Regards

### DIFF
--- a/russh-keys/src/backend_rust.rs
+++ b/russh-keys/src/backend_rust.rs
@@ -155,13 +155,11 @@ impl<'a> From<&RsaPrivate> for protocol::RsaPublicKey<'a> {
     }
 }
 
-impl TryFrom<&RsaPrivate> for RsaPublic {
-    type Error = Error;
-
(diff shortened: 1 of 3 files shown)


nicmr:
taking a look at the compile error

nicmr:
Sorry, I totally missed to check with the openSSL backend, which is why this has to be fallible, and I only use the pure rust implementation locally. At least I understand the reason know. :facepalm:

PR #368, opened 2024-10-30
Use ssh-key instead of local key handling

Eugeny:


Nathy-bajo:
Looks good to me! @Eugeny 😁

PR #374, opened 2024-11-13
Derive Debug where possible

qsantos:
I mostly wanted to `dbg!` the server's `Session`, but I took the opportunity to add `#[derive(Debug)]` wherever I could.


Eugeny:
A lot of these are empty structs and structs with security relevant fields that should not end up in some forgotten `dbg!` output. I could merge a handmade `Debug` impl with just the relevant fields for Session but not this.

qsantos:
I have excluded the private keys from the `Debug` implementation of the server `Session`, but I probably missed some other secrets.

Issue #373, opened 2024-11-12
Handle client disconnections


EpicEric:
Currently, there's only handling for sessions errors - and even in this case, there's no way to refer to the handler itself, only to the error. In cases like the library-provided examples `echoserver` and `ratatui_*`, this leads to leaks in the form of `Arc<Mutex<HashMap<...>>>` entries, which is troublesome for any real application.

From what I've managed to hunt in the source code so far, it only seems to [log the disconnection](https://github.com/Eugeny/russh/blob/ad2403ba4ea513b743406cbbb076278ccae98029/russh/src/server/mod.rs#L599-L604) without doing much else. Ideally, there should be a `client_disconnected` callback that takes the `Server::Handler` instance in order to handle cleanup, for example.

Patryk27:
fwiw, I'm using `impl Drop` to know when a client disconnects and it seems to work pretty reliably.

Eugeny:
Correct - simply do your cleanup in `drop()` as the handler gets dropped when the session is closed.

PR #376, opened 2024-11-14
impl Drop for server examples

EpicEric:
In #373 I raised my confusion about the lack of a disconnection mechanism in the source code and the examples, but it was pointed out that `Drop` is the intended way of cleaning up resources.

This PR adds a Drop implementation for three of the server examples, so that users that refer to these won't be confused about how to implement their own cleanup.


Eugeny:
Thank you! 

@all-contributors add @EpicEric for code

PR #381, opened 2024-11-17
Document how to reply to channel requests

qsantos:
See https://github.com/Eugeny/russh/issues/156.

Using Rust typing system would be less error-prone. Maybe the [`pty-req` case in `server_read_authenticated`](https://github.com/Eugeny/russh/blob/7dee93524e77cf78229eeacb3ccaaf5f77a9036b/russh/src/server/encrypted.rs#L881-L905) could handle it in two ways:

- call `Handle::channel_success` or `Handle::channel_failure` depending on the value returned by the handler;
- always call Handle::channel_success` after calling the handler, assuming that, if the user calls `Handle::channel_failure` in the handler, it will have priority; feels very hacky though.

### DIFF
--- a/russh/src/server/mod.rs
+++ b/russh/src/server/mod.rs
@@ -396,6 +396,27 @@ pub trait Handler: Sized {
 
     /// The client requests a pseudo-terminal with the given
     /// specifications.
+    ///
+    /// **Note:** Success or failure should be communicated to the client by calling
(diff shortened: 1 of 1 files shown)


qsantos:
Regarding the second part of my comment, I just realized that this is what https://github.com/Eugeny/russh/pull/348 tries to do.

Eugeny:
Thanks! @all-contributors add @qsantos for docs

PR #391, opened 2024-11-23
Implement From<&str> and From<&[u8]> for CryptoVec

joshka:
So that `session.data(channel, "foo".into())` works.

### DIFF
--- a/cryptovec/src/cryptovec.rs
+++ b/cryptovec/src/cryptovec.rs
@@ -368,25 +368,18 @@ impl Drop for CryptoVec {
     }
 }
 
-// DocTests cannot be run on with wasm_bindgen_test
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


joshka:
I wanted to some simple tests for this, but wasn't sure why the test module was platform gated. Pinging @irvingoujAtDevolution for some insight on the rationale.

irvingoujAtDevolution:
If I recall correctly, it was not there when I first created the test. `Vec<u8>` was a replacement for some libc dependent code under WASM.

irvingoujAtDevolution:
You can safely remove it in this case, at least for these two tests.

joshka:
> You can safely remove it in this case, at least for these two for now.

Sounds good: 5980e71..72cd436

Eugeny:
LGTM, thanks!

@all-contributors add @joshka for code

PR #397, opened 2024-11-29
Fix ChannelStream's channel not being closed

### LINKED ISSUE #396: ChannelStream leaks in remote port forwarding
When remote port forwarding for multiple connections, memory usage increases linearly, eventually leading to OOM in my use case.

You can see a demo here, which consumes more and more memory (can be verified in `top`): https://github.com/EpicEric/russh-stress-test

This appears to be related to dangling connections, as restarting either the server or the remote forwarded client causes the memory usage to drop down on the opposite side. The data transmitted through the stream also doesn't affect the memory increase, as the example shows.

EpicEric:
Closes #396

I'll test this in my project later today

### DIFF
--- a/russh/src/channels/io/rx.rs
+++ b/russh/src/channels/io/rx.rs
@@ -89,3 +89,16 @@ where
         }
     }
 }
+
+impl<'i, S> Drop for ChannelRx<'i, S>
+where
+    S: From<(ChannelId, ChannelMsg)>,
+{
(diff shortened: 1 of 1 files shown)


EpicEric:
I'm not sure why the test is failing. It seems to fail on current main as well.

EpicEric:
I've tried out this change and it works like a charm, with no more memory increases.

PR #398, opened 2024-11-30
Expose authenticated username for session

zaczkows:
- this allows to modify sftp server behaviour depending on logged user


Eugeny:
It's trivial to save the username in your `Handler` implementation. Besides, `auth_user` is set even before a successful authentication and should not be relied on in this situation.

PR #399, opened 2024-11-30
Add option to directly load openssh private key file in pem format

zaczkows:
- although this can be done manually, it's just more convenient to have it out-of-the-box


Eugeny:
Isn't this already covered by https://docs.rs/russh-keys/0.49.0/russh_keys/fn.load_secret_key.html ?

zaczkows:
The functionality itself - yes, but it requires a file. It would be more convenient to have a function that can load directly a file content.

Eugeny:
I see. This function is so trivial however that I wouldn't consider including it in the library itself. Appreciate the PR though :v:

PR #400, opened 2024-11-30
Fix improper RSA key conversion from `ssh_key` crate

EpicEric:
`rsa` 0.9.7 introduces an error when using a key to sign (likely due to overeager validation during conversion). You can see the changes here: https://diff.rs/rsa/0.9.6/0.9.7/

This commit fixes the version back to 0.9.6 as a quick fix, although this may not be the desired solution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ futures = "0.3"
 hmac = "0.12"
 log = "0.4"
 rand = "0.8"
-rsa = "0.9"
+rsa = "=0.9.6"
 sha1 = { version = "0.10", features = ["oid"] }
 sha2 = { version = "0.10", features = ["oid"] }
 signature = "2.2"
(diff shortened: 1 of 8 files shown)


EpicEric:
Tracking issue: https://github.com/RustCrypto/RSA/issues/463

EpicEric:
I've updated this with a proper workaround while https://github.com/RustCrypto/SSH/pull/318 is not merged.

Eugeny:
Thank you for investigating this!

Cargo.toml
```diff
@@ -26,7 +26,7 @@ futures = "0.3"
 hmac = "0.12"
 log = "0.4"
 rand = "0.8"
-rsa = "0.9"
+rsa = "=0.9.6"
```
dignifiedquire, line 29: I would recommend not pinning to `0.9.6` to make sure this issue doesn't occur in other code paths, given that you have a workaround now

Issue #401, opened 2024-12-01
100% CPU usage after connection lost while sending large amount of data


PonyPC:
100% CPU usage after connection lost while sending large amount of data

The issue is in https://github.com/Eugeny/russh/blob/ff95994945db593115dd0d663f8c41ecba945b9f/russh/src/channels/io/tx.rs#L64

It won't triggle connection lost event.

EpicEric:
Shouldn't this trigger an `Err(BrokenPipe)`? Or am I missing some other way that would lead to the condition being reached?

PonyPC:
Nothing triggered. If you add a check like:
```
            if self.sender.is_closed() {
                return Poll::Ready((ChannelMsg::Eof, 0));
            }
```
It only mitigates the tcp las a connection lost event. But when you try it such as block the connection by using a firewall realtime rules, the CPU usage is still high until tcp timeout error. As it not sufficient on dealing with sending data, is there any other ways to eliminate this situation.

PonyPC:
Nothing triggered. If you add a check like:
```
            if self.sender.is_closed() {
                return Poll::Ready((ChannelMsg::Eof, 0));
            }
```
It only mitigates the tcp has a connection lost event. But when you try it such as blocking the connection by using a firewall realtime rules, the CPU usage is still high until tcp timeout error. As it not sufficient on dealing with sending data, is there any other ways to eliminate this situation?

PR #407, opened 2024-12-04
Attempt to remove busywait in ChannelTx

EpicEric:
Some progress on #401. This passes the `poll_mk_msg` waker along with the window size in order to avoid busy-waiting for changes. This is currently untested, hence the draft PR.

### DIFF
--- a/russh/src/channels/io/tx.rs
+++ b/russh/src/channels/io/tx.rs
@@ -9,8 +9,7 @@ use tokio::sync::mpsc::error::SendError;
 use tokio::sync::mpsc::{self, OwnedPermit};
 use tokio::sync::{Mutex, OwnedMutexGuard};
 
-use super::ChannelMsg;
-use crate::{ChannelId, CryptoVec};
+use crate::{channels::WindowSize, ChannelId, ChannelMsg, CryptoVec};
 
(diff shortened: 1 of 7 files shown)


EpicEric:
@PonyPC Could you verify if this PR works for you?

russh/src/channels/io/tx.rs
```diff
@@ -58,11 +57,13 @@ where
         self.window_size_fut.take();
 
         let writable = (self.max_packet_size)
-            .min(*window_size)
+            .min(window_size.get())
             .min(buf.len() as u32) as usize;
         if writable == 0 {
-            // TODO fix this busywait
-            cx.waker().wake_by_ref();
+            if buf.is_empty() || self.sender.is_closed() {
```
EpicEric, line 63: I am not sure if `buf.is_empty()` should lead to the channel being closed, but maybe the condition above should only be reached when the window_size is zero?

EpicEric:
Mentioning the #197 PR here as it seems to have a more nuanced approach to the optimal solution.

Eugeny:
@EpicEric I've made a very rough attempt at this in #408 (not tested yet) - I'm not very good at low level async and could use a review
It's using an `mpsc::watch` as a notification channel for window size changes and waking up all senders at once when the window size changes - this wastes a poll if the window size is not enough for all senders but that should be fine.

PR #408, opened 2024-12-04
An attempt at #401 - removing TX busywait

Eugeny:


russh/src/channels/io/tx.rs
```diff
@@ -35,50 +76,72 @@ where
...
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<(ChannelMsg, NonZero<usize>), watch::error::RecvError>> {
+        let writable = match ready!(self.poll_writable(cx, buf.len())) {
+            Ok(w) => w,
+            Err(e) => return Poll::Ready(Err(e)),
+        };
+
+        let mut data = CryptoVec::new_zeroed(writable.into());
+        #[allow(clippy::indexing_slicing)] // Clamped to maximum `buf.len()` with `.min`
```
EpicEric, line 135:
```suggestion
        #[allow(clippy::indexing_slicing)] // Clamped to maximum `buf.len()` with `.poll_writable`
```

EpicEric (commented):
I don't see a way where the changes to the window size are communicated back to the watchers.

Otherwise this might work (I'm also not very familiar with low-level async), although something like [`Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) might be a better fit as it only lets in one task at a time, which is already the case given the Mutex for the window size.

Eugeny:
Dang, I wasn't testing it correctly in the first place, thanks for the fix!
I've wrapped the window size updates in a fn so that the notifier fires after the update has already happened just in case there might be an await point in the future between the two.

I'm also wondering whether `notify_all` is needed instead of `notify_one` as a freshly awakened writer can be dropped before it is polled again.

EpicEric:
Good point, that seems like the better solution.

EpicEric:
Hmm, I'm not sure how Russh actually handles window resizes. I have the following test case:

```rust
let mut writer = ch.make_writer();
let fut = tokio::spawn(async move {
    let mut buf = [0u8; 1024 * 64];
    thread_rng().try_fill_bytes(&mut buf).unwrap();
    assert!(writer_2.write_all(&buf).await.is_ok());
});
assert!(tokio::try_join!(fut).is_ok());  // Removed other futures for clarity
```

But it seems to get stuck when invoking `write_all`, as the window resize is never invoked. Changing every call to `notify_watchers` also doesn't fix the issue.

Eugeny:
@EpicEric is there anything you'd like to see added here?

EpicEric:
This looks good to me.

PR #411, opened 2024-12-05
Further improvements to WIP #408

EpicEric:
- Guard against empty buffer
- Fix tests not checking for panics
- Add test_channel_window_size with multiple writers


Eugeny:
Looking good and I appreciate the test :+1: :+1:

PR #415, opened 2024-12-16
Add `authentication_banner` method to server::Handler

### LINKED ISSUE #372: Feature request: support setting banner on each connection
I'd like to set a custom, unique banner for every incoming SSH connection. However, it appears that [the banner is the same for all connections](https://docs.rs/russh/latest/russh/server/struct.Config.html#structfield.auth_banner). It would be great if this was possible!

EpicEric:
This replaces the configuration option `auth_banner`, making this a breaking change.

Closes #372.


Eugeny:
Looking good, thank you!

PR #419, opened 2024-12-19
bump patch version of dependency `ssh-key`

eric-seppanen:
The feature `p521` doesn't exist in versions of `ssh-key` older than 0.6.2.

This causes build failures if `russh` is introduced into a cargo workspace that already contains a dependency on e.g. `ssh-key 0.6.0`; since nothing in the `russh` `Cargo.toml` was specifying an update was required Cargo does not know to update to a newer version.


Eugeny:
Already bumped in main

PR #423, opened 2024-12-20
replace tempdir with tempfile

eric-seppanen:
`tempdir` has been abandoned, last published 6 years ago. (RUSTSEC-2018-0017). And some of its dependencies have known security issues (RUSTSEC-2023-0018).

The recommended replacement is the `tempfile` crate.

Most uses of `tempdir` were in tests, but one was in the example in `russh-keys`-- it's probably not a good idea for examples to include code that would bring in dependencies with security issues.

This resolves two advisories reported by `cargo-deny`. It might be a good idea to run `cargo deny` periodically to catch issues with dependencies (probably not on the PR path since it will spontaneously start failing when a new advisory lands).

### DIFF
--- a/russh-keys/Cargo.toml
+++ b/russh-keys/Cargo.toml
@@ -82,5 +82,5 @@ pageant = { version = "0.0.1-beta.3", path = "../pageant" }
 
 [dev-dependencies]
 env_logger = "0.11"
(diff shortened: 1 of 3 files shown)


Eugeny:
Thanks!

PR #424, opened 2024-12-20
feature-gate `des` dependency

eric-seppanen:
Move the `des` dependency behind a cargo feature that is disabled by default, both to avoid security concerns about dependencies (an auditor may not know which ciphers are enabled at runtime), as well as reduce the number of dependencies downstream projects need to build.

This addresses one of the issues in #420.


Eugeny:
LGTM :+1:

Eugeny:
@all-contributors add @eric-seppanen for code

PR #428, opened 2024-12-21
bump dependency versions to the minimum version that compiles.

ericseppanen:
The dependency versions in Cargo.toml were inaccurate in places, because they allowed versions that don't actually work-- crates were enabling Cargo features that don't exist in those versions, or methods/traits/etc that didn't exist in the code.

This allows a tool like `cargo minimal-versions` to verify that the code compiles even when using the oldest possible dependency versions, but is also a useful step toward verifying that the crates can be built on older toolchains.

Replaces #419.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -80,3 +80,14 @@ jobs:
         cargo test --verbose --all-features
       env:
         RUST_BACKTRACE: 1
+
+  Minimal-versions:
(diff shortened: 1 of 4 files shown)


Eugeny:
Looking good! I've added a CI job to run minimal-versions as well

Eugeny:
@all-contributors add @ericseppanen for code

Issue #444, opened 2025-01-12
RSA tests fail when running on systems with a cryptographic policy


nightmared:
Hello,

The tests `test_client_agent_openssh_rsa` and `test_client_agent_rsa` fail when running on a system with a cryptographic policy (which is enabled by default on RedHat-like distros, like Fedora).

By default, we get:
```
sthoby@tolkien:~/dev/russh$ cargo test -p russh-keys
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests src/lib.rs (target/debug/deps/russh_keys-730ace141644eff2)

running 28 tests
test test::test_decode_openssh_p384_secret_key ... ok
test test::test_decode_openssh_p256_secret_key ... ok
test known_hosts::test::test_check_known_hosts ... ok
test test::test_decode_openssh_p521_secret_key ... ok
test test::test_client_agent_openssh_rsa ... FAILED
test test::test_decode_pkcs8_rsa_secret_key ... ok
test test::test_decode_pkcs8_p256_secret_key ... ok
test test::test_decode_rfc8410_ed25519_private_only_key ... ok
test test::test_decode_rfc8410_ed25519_private_public_key ... ok
test test::test_fingerprint ... ok
test test::test_decode_rsa_secret_key ... ok
test test::test_gpg ... ok
test test::test_nikao ... ok
test test::test_parse_p256_public_key ... ok
test test::test_parse_p384_public_key ... ok
test test::test_parse_p521_public_key ... ok
test test::test_o01eg ... ok
test test::test_pkcs8 ... ok
test test::test_srhb ... ok
test format::tests::test_ec_private_key ... ok
test test::test_client_agent_rsa ... FAILED
test test::test_pkcs8_encrypted ... ok
test test::test_decode_pkcs8_p521_secret_key ... ok
test test::test_agent ... ok
test test::test_decode_pkcs8_p384_secret_key ... ok
test test::test_decode_ed25519_secret_key ... ok
test test::test_decode_ed25519_aesctr_secret_key ... ok
test test::test_client_agent_ed25519 ... ok

failures:

---- test::test_client_agent_openssh_rsa stdout ----
thread 'test::test_client_agent_openssh_rsa' panicked at russh-keys/src/lib.rs:859:58:
called `Result::unwrap()` on an `Err` value: AgentFailure
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::test_client_agent_rsa stdout ----
[...]

PR #449, opened 2025-01-14
Allow setting hash algorithm to use for signing requests of SSH agent

wiktor-k:
This is a suggestion PR that I've tested locally. I'm happy to adjust it in any way you see fit.

@nightmared would it be possible for you to test it, too? (just like in the ticket you've described).

Fixes: https://github.com/Eugeny/russh/issues/444
Fixes: https://github.com/Eugeny/russh/issues/445

Note that since I'm using my own agent server implementation and upstream ssh-key has not released a version with some critical fixes such as https://github.com/RustCrypto/SSH/pull/263 making signatures with SHA-1 is totally broken :( so I'd be really happy if something like this has been merged. Thanks for your time! :bow:

### DIFF
--- a/russh/src/auth.rs
+++ b/russh/src/auth.rs
@@ -18,7 +18,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use ssh_key::{Certificate, PrivateKey};
+use ssh_key::{Certificate, HashAlg, PrivateKey};
 use thiserror::Error;
 use tokio::io::{AsyncRead, AsyncWrite};
 
@@ -154,6 +154,7 @@ pub trait Signer: Sized {
     async fn auth_publickey_sign(
         &mut self,
         key: &ssh_key::PublicKey,
+        hash_alg: Option<HashAlg>,
         to_sign: CryptoVec,
     ) -> Result<CryptoVec, Self::Error>;
 }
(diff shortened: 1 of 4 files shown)


Eugeny:
I think a more flexible solution would be to update the `Signer` trait to accept an `Option<HashAlg>` parameter, similar to how `PrivateKeyWithHashAlg` is used for direct key auth. If you have time to implement this change, please do, otherwise I'll merge it as-is and do it later, thanks!

It will need to be updated later to accept an enum of a _public key OR certificate_ later for OpenSSH certificate keys agent auth.

wiktor-k:
Thanks for the feedback! I'll try to implement your idea tomorrow and see how far I go...

nightmared:
This PR does fix the test (and would close #444)!

However, I'm not convinced hardcoding SHA512 (instead of leaving the option to specify the algorithm as @Eugeny suggested) is the best choice here. E.g., what would happen when talking to old systems/SSH servers that may not support it? I fear this could introduce regressions.

Eugeny:
(sorry for the conflict, client.rs got moved from `russh-keys` to `russh/src/keys `on main)

wiktor-k:
> (sorry for the conflict, client.rs got moved from `russh-keys/src` to `russh/src/keys `on main)

Fortunately git took care of that seamlessly when rebasing :sweat_smile: 

Okay, I've passed the `hash_alg` throughout the code and verified that this PR (still) works with my other code... phew. (Note that I've adjusted only signing with the agent, the thing that I use, maybe some other parts need adjusting, too? :thinking: )

See if that's more like what you meant and if not I can re-adjust.

Thanks for your time! :wave:

Eugeny:
LGTM, thank you!

PR #454, opened 2025-01-26
Add client demo that implement open direct tcpip.

handewo:
This demo works like `ssh -L <bind_addr:forward_addr> -N <host>`
Known issue: forward data sometimes incorrect since not handle `ChannelMsg::WindowAdjusted`, but I think it's not important as demo.


handewo:
> Thank you for the PR. I'd rather not overload the new example with certificate handling. If you'd like, you can instead add the certificates support to client_exec_simple

No problem. I've added.

Eugeny:
Could you please also remove the usage of `async_trait` to fix the build? It's now optional in the crate and shouldn't be used in the examples

handewo:
> Could you please also remove the usage of `async_trait` to fix the build? It's now optional in the crate and shouldn't be used in the examples

Done.

Eugeny:
Thanks! @all-contributors add @handewo for code

PR #461, opened 2025-02-07
Fix reading on a closed channel throw `UnexpectedEof` error

uuhan:
Stop reading on a closed channel and stop it's keepalive timer.

This may cause an Error after a session closed:

```txt
SshError(IO(Custom { kind: UnexpectedEof, error: "early eof" }))
```

I encountered in my [workhorse](https://github.com/uuhan/workhorse/blob/7b37aa801e04822d875b466ddaf746e04fddcf5f/horsed/src/ssh/mod.rs#L992) project, and also find a related issue: https://github.com/Eugeny/russh/issues/253

After a long trace in the source code, found that this is caused by 

1. read data after channel closed:

https://github.com/Eugeny/russh/blob/89fed88265b402e7d505025c42f4e19ab12201f3/russh/src/server/mod.rs#L903
https://github.com/Eugeny/russh/blob/89fed88265b402e7d505025c42f4e19ab12201f3/russh/src/cipher/mod.rs#L254
(insufficient data to be read)
this caused the "early eof" issue.

2. still doing keepalive after channel closed:

https://github.com/Eugeny/russh/blob/89fed88265b402e7d505025c42f4e19ab12201f3/russh/src/server/session.rs#L525

this caused the "broken pipe" issue.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -474,13 +474,15 @@ impl Session {
         let reading = start_reading(stream_read, buffer, opening_cipher);
         pin!(reading);
         let mut is_reading = None;
+        let mut channel_closed = false;
 
         #[allow(clippy::panic)] // false positive in macro
         while !self.common.disconnected {
             self.common.received_data = false;
             let mut sent_keepalive = false;
             tokio::select! {
-                r = &mut reading => {
+                r = &mut reading, if !channel_closed => {
+                    debug!("stream reading");
                     let (stream_read, mut buffer, mut opening_cipher) = match r {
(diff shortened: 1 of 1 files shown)


Eugeny:
This change just kills the session as soon as the first channel closes.
In the SSH protocol, it's legal to close one or even all channels and still keep the connection running, including sending new global requests (e.g. keepalives) and opening new channels later.

Unfortunately there is no way for the server to distinguish between an accidental `early eof` and the client closing the connection itself without sending a clean `MSG_DISCONNECT` first.

So in the end, it would be up to you as a library consumer to decide whether you want to propagate `early off` errors to your user or just treat them all as normal disconnects.

uuhan:
Got it.

PR #465, opened 2025-02-09
Add a function to send ExitStatus message to Channel

procr1337:
This is a non-breaking change which adds a way to send ExitStatus over a channel. This makes it possible to write simple command/shell handlers which consume a Channel<Msg> and perform some shell interaction. Such handlers can then be spawned in response to an `exec_request` or `shell_request`.

Using the &mut Session to try and send the ExitStatus message instead is not easy for longer living interactions (that don't immediately return).


Eugeny:
Thank you!

PR #471, opened 2025-02-21
Cryptovec/Windows: Add reference counting per Page, improve error-msg

amtelekom:
Hi,

the current CryptoVec implementation on Windows unlocks memory pages too early if the backing stores of multiple CryptoVecs are located in the same page.

This is because [VirtualLock](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtuallock)/[Unlock](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualunlock) do not implement reference-counting on the OS side:

> This function is not like the GlobalLock or LocalLock function in that it does not increment a lock count and translate a handle into a pointer. There is no lock count for virtual pages, so multiple calls to the VirtualUnlock function are never required to unlock a region of pages.

This PR adds userspace lock counting, so only once the last CryptoVec has released it's memory from a Page, the Page is unlocked.

### DIFF
--- a/cryptovec/src/platform/windows.rs
+++ b/cryptovec/src/platform/windows.rs
@@ -1,24 +1,99 @@
-use libc::c_void;
+use std::ffi::c_void;
+use std::sync::{Mutex, OnceLock};
 use winapi::shared::basetsd::SIZE_T;
 use winapi::shared::minwindef::LPVOID;
+use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::memoryapi::{VirtualLock, VirtualUnlock};
+use winapi::um::sysinfoapi::{GetNativeSystemInfo, SYSTEM_INFO};
 
 use super::MemoryLockError;
 
+// To correctly lock/unlock memory, we need to know the pagesize:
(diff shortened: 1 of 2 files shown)


cryptovec/src/platform/windows.rs
```diff
@@ -1,24 +1,99 @@
...
+        match locked_pages.binary_search_by_key(&page_idx, |pl| pl.page_idx) {
+            Ok(vec_idx) => {
+                #[allow(clippy::indexing_slicing)] // binary_search Ok() is always valid index
+                {
+                    locked_pages[vec_idx].lock_counter += 1;
+                }
+            }
+            Err(vec_idx) => {
+                lock_page(page_idx)?;
+                locked_pages.insert(
```
Eugeny, line 77: Inserting into a `Vec` negates the savings from `binary_search_by_key` here. Something like a `BTreeMap` would be a better choice for storing counters.

Eugeny (changes requested):

PR #472, opened 2025-02-21
Pageant: Implement NamedPipes-based stream

amtelekom:
Adds an implementation of the newer NamedPipes-based access to the Pageant Stream.

Right now, it can be switched to via feature, i.e. the binary crate imports

```
pageant = { version = "0.0.3", features = ["namedpipes"] }
```

to enable it.

This is a breaking change both on the API-Surface of the Pageant crate (new needs to be async + have error handling), as well as in russh (also to introduce a `Result` return type).

Alternatively to this implementation, we could also just offer the generation of the Named-Pipe path in this crate and let users create a NamedPipe-Backed agent using the existing code in Russh, although this version seemed more comfortable to me.


pageant/src/lib.rs
```diff
@@ -11,8 +11,14 @@
     clippy::panic
 )]
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "wmmessage", not(feature = "namedpipes")))]
```
Eugeny, line 14: Thanks a lot!

The two implementations should definitely not be mutually exclusive. They could be both exported together, with neither being a "default", e.g.:

```rust
#[cfg(all(windows, feature = "wmmessage"))]
pub use pageant_impl::PageantStream as PageantWmMessageStream;

#[cfg(all(windows, feature = "namedpipes"))]
pub use pageant_impl_namedpipes::PageantStream as PageantNamedPipeStream;
```

I'd really like to see a common trait for both of them since it would be weird to have to choose a specific implementation at compile time unless you know exactly which PuTTY version to expect in advance

Eugeny (changes requested):

pageant/src/lib.rs
Rondom: There is a cost to maintaining both and why would one want to run an old pre-2020 Putty Pageant with potential security bugs in 2025? I think a compile time decision is good enough, and I would even evaluate to eventually remove the  WM_MESSAGE method.

Eugeny: Because in reality, when you're writing an SSH client, you don't control what version of Putty the user might have installed.

Eugeny: P.S. I don't mind doing it myself, it will just take a bit until I have time to work on it.

amtelekom:
I've added an enum that can hold either Transport, and allows for trying both mechanisms if compiled in via the `new` on the enum. Since we already just rely on `ReadStream` / `WriteStream` traits, adding a specific one for Pageant impls wasn't necessary.
Via features, you can still compile out specific variants, but by default, both impls are now included and tried.

The breaking API changes mentioned in the initial PR message remain - I've bumped the Pageant crate version, but didn't include one for the russh crate in the PR, since I'm not sure how you'd categorize the change.

Eugeny:
Thank you for coming back to this! LGTM now

PR #473, opened 2025-02-25
Add crate features to enable insecure ciphers and algos

EdMcBane:
Adds a crate feature to optionally enable support for insecure ciphers and algos such as RSA keys shorter than 2048 bits and DSA keys in dependency `ssh-keys`.


Eugeny:
Thanks!

Eugeny:
@all-contributors add @EdMcBane for code

PR #477, opened 2025-02-27
Optimize examples/client_open_direct_tcpip.rs

handewo:
I think ChannelMsg::WindowAdjusted doesn't required processing by user.


Eugeny:
Thanks!

PR #481, opened 2025-03-03
build: fix compilation failure on apple targets

Itsusinn:


Itsusinn:
https://github.com/Eugeny/russh/issues/480

PR #482, opened 2025-03-03
Introduce Channel::split() to allow splitting a channel into a read half and a write half

psychon:
Hi there,

here is my attempt at fixing https://github.com/Eugeny/russh/issues/476. A tiny bit more details can be found in the commit messages.

One thing that I did not mention there: [The original suggestion](https://github.com/Eugeny/russh/issues/476#issuecomment-2687214367) was:

> To avoid duplication, Channel itself could just itself be a simple wrapper for the two halves that uses delegate! to forward all methods.

I did not manage to figure out how to make doc comments work together with `delegate!`, so I implemented the delegation by hand (and did mistakes in the process, which were found thanks to rustc's "unused parameter" warnings).

### DIFF
--- a/russh/src/channels/io/rx.rs
+++ b/russh/src/channels/io/rx.rs
@@ -42,7 +42,7 @@ where
     ) -> Poll<io::Result<()>> {
         let (msg, mut idx) = match self.buffer.take() {
             Some(msg) => msg,
-            None => match ready!(self.channel.as_mut().receiver.poll_recv(cx)) {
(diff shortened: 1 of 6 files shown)


Eugeny:
Looking good, thanks for the PR!

Eugeny:
@all-contributors add @psychon for code

MingweiSamuel:
Is there a reason `make_writer_ext` is private on the write half?

PR #488, opened 2025-03-07
Remove unused dependencies

psychon:
I ran cargo-udeps on this workspace (on a Linux machine) and looked at the results.

- pageant is an empty crate on non-Windows, so all of its dependencies are Windows-only
- russh uses tokio-stream only in a doc-test. For this, dev-dependencies suffices. I do not see why this is "not wasm32", but it already was this way before, so I leave it there.
- russh-util does not seem to use futures-executor or static_assertions and apparently never did (??). These were added in commit d03df31af. In this commit, I am dropping them.
- russh-util uses chrono only on wasm

### DIFF
--- a/pageant/Cargo.toml
+++ b/pageant/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/warp-tech/russh"
 version = "0.0.2"
 rust-version = "1.75"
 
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you!

Issue #490, opened 2025-03-12
undeclared crate or module `libc` in 0.51.1 on windows


VladTheJunior:
after upgrading to 0.51.1 from 0.50.4 I can't build my app 

```
error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> C:\Users\vladt\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\russh-cryptovec-0.51.0\src\platform\windows.rs:92:9
   |
92 |         libc::memset(ptr as *mut c_void, value, size);
   |         ^^^^ use of undeclared crate or module `libc`
For more information about this error, try `rustc --explain E0433`.
error: could not compile `russh-cryptovec` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

psychon:
I guess this comes from #483. Is there a reason why this PR cannot do what that PR does and replace `libc::memset` with `nix::libc::memset`? (Assuming this actually works...?)

https://github.com/Eugeny/russh/pull/483/files#diff-125248874f5781f71e028308212b0320e5f067a5c20bc858b8d91b5c3c4647eaL28-R31

Eugeny:
russh-cryptovec@0.50.1 is released

PR #491, opened 2025-03-12
fix: #490 add libc for windows

Caisin:


Eugeny:
I was a bit faster and it's fixed in the latest cryptovec, but thank you anyway!

PR #495, opened 2025-03-17
Handle unexpected EOF as expected in session closure

lgmugnier:
Fixes: https://github.com/Eugeny/russh/issues/253

My session was hitting the following error after receiving a `DISCONNECT` message from the client.
```
IO(Custom { kind: UnexpectedEof, error: "early eof" })
```

On closer inspection (and thanks to the issue), I realized it was coming from the shutdown process of the session. I believe the intention here is to break when we hit eof (read 0 bytes):
https://github.com/Eugeny/russh/blob/89fed88265b402e7d505025c42f4e19ab12201f3/russh/src/server/session.rs#L650
but because of this line in the called read function
https://github.com/Eugeny/russh/blob/89fed88265b402e7d505025c42f4e19ab12201f3/russh/src/cipher/mod.rs#L254
reading 0 bytes will always be an UnexpectedEOF error. This might be good to fully propagate the error other places in the code, but here it makes sense to catch it and process it as an expected EOF.

I kept in the n == 0 check but I doubt we'd ever see that case fire since it would mean that we received an empty ssh message and I'd imagine the reason this loop exists is to make sure we're not missing any actual messages while we wait for everything to shut down.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -652,10 +652,10 @@ impl Session {
                 Ok((0, _, _, _)) => break,
                 Ok((_, r, b, opening_cipher)) => {
                     is_reading = Some((r, b, opening_cipher));
-                },
-                // read_exact returns this error if it reads 0 bytes, this is an expected eof
-                Err(Error::IO(ref e)) if e.kind() == ErrorKind::UnexpectedEof=> break,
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! I've just corrected the comment

PR #505, opened 2025-04-01
bump russh-sftp

AspectUnk:
added native async-trait support https://github.com/AspectUnk/russh-sftp/issues/62


Eugeny:
Thank you!

PR #510, opened 2025-04-03
Can send debug messages

pgrange:
Thank you for russh 😻

I needed to send _debug_ messages. I implemented the same way disconnect is implemented as they both are session global messages.

Looking at the test suites, I did not find any appropriate place to add a test for that.

One can observe the debug messages sent by a server, for instance, by activating the `-v` option of ssh client.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -752,6 +752,33 @@ impl Session {
         self.common.disconnect(reason, description, language_tag)
     }
 
+    /// Sends a debug message to the client.
+    ///
+    /// Debug messages are intended for debugging purposes and may be
+    /// optionally displayed by the client, depending on the
+    /// `always_display` flag and client configuration.
+    ///
+    /// # Parameters
+    ///
+    /// - `always_display`: If `true`, the client is encouraged to
+    ///   display the message regardless of user preferences.
+    /// - `message`: The debug message to be sent.
+    /// - `language_tag`: The language tag of the message.
+    ///
+    /// # Notes
+    ///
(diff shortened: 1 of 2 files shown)


russh/src/server/session.rs
```diff
@@ -752,6 +752,15 @@ impl Session {
         self.common.disconnect(reason, description, language_tag)
     }
 
+    pub fn debug(
```
Eugeny, line 755: add a docstring

russh/src/lib.rs
```diff
@@ -114,7 +114,7 @@ pub mod mac;
 
 pub mod keys;
 
-mod msg;
+pub mod msg;
```
Eugeny, line 117: revert

russh/src/session.rs
```diff
@@ -189,6 +189,28 @@ impl<C> CommonSession<C> {
         Ok(())
     }
 
+    pub fn debug(
+            &mut self,
```
Eugeny, line 193: bad indent

Eugeny (changes requested):
looking good - just a few nitpicks

russh/src/server/session.rs
pgrange: Docstring added 👍

russh/src/lib.rs
pgrange: Fixed, sorry about that.

Eugeny:
@all-contributors add @pgrange for code

Eugeny:
Thanks!

PR #511, opened 2025-04-04
Add Handle::send_keepalive()

psychon:
In this commit, I add a function to russh's client Handle that allows to manually send a keepalive message to the server.

Fixes: https://github.com/Eugeny/russh/issues/507


Eugeny:
Thank you!

PR #518, opened 2025-05-21
Fail when accept returns an Error

pgrange:
For instance, if the service can not accept any new connection, instead of silently terminating, it will return the underlying error:
```
Os { code: 24, kind: Uncategorized, message: "Too many open files" }
```


Eugeny:
Thank you!

PR #524, opened 2025-06-06
Fix: Dead loop when client receives messages

wyhaya:
```rust
use russh::Error;
use russh::client::{Config, Handler, connect};
use russh::keys::PublicKey;
use russh::keys::agent::client::AgentClient;
use std::sync::Arc;
use std::time::Duration;
use tokio::time::sleep;

#[tokio::main]
async fn main() {
    test().await;
    sleep(Duration::from_secs(100)).await;
}

async fn test() {
    let client = Client {};
    let config = Arc::new(Config::default());
    let mut session = connect(config, ("127.0.0.1", 22), client).await.unwrap();

    let path = r#"/Users/.../Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock"#;
    let mut agent = AgentClient::connect_uds(path).await.unwrap();
    let mut keys = agent.request_identities().await.unwrap();
    let key = keys.remove(0);
    let rst = session
        .authenticate_publickey_with("root", key, None, &mut agent)
        .await;
    dbg!(rst.err());
}

struct Client {}

impl Handler for Client {
    type Error = Error;
    async fn check_server_key(&mut self, _: &PublicKey) -> Result<bool, Self::Error> {
        Ok(true)
    }
}
```

This is a simple example that uses the `1Password SSH Agent` for authentication. Once I reject the authorization on `1Password`, the program enters an `infinite loop`, causing the CPU usage to reach 100%.

I am not familiar with the SSH protocol; I just happened to discover this issue. Please feel free to point out any side effects if there are any!

Thanks, russh.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -14,6 +14,7 @@
 //
 use std::cell::RefCell;
 use std::convert::TryInto;
+use std::io::Error as IoError;
 use std::ops::Deref;
 use std::str::FromStr;
 
@@ -239,6 +240,12 @@ impl Session {
                                     self.common.buffer = loop {
                                         match self.receiver.recv().await {
                                             Some(Msg::Signed { data }) => break data,
+                                            None => {
+                                                return Err(crate::Error::IO(IoError::other(
+                                                    "Failed to receive message",
+                                                ))
+                                                .into());
+                                            }
                                             _ => {}
                                         }
                                     };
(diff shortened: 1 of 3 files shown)


wyhaya:
By searching for `self.receiver.recv()`, I found that there are also unhandled `None` cases in `russh/src/client/mod.rs`. I am not sure whether it is necessary to modify it together or if there is no need to modify it at all.

Eugeny:
Thank you and sorry for the delay - feel free to add handling for other cases. 
Please also add a custom `RecvError` error variant and use that instead of `Disconnect`

Eugeny:
Thanks!

Eugeny:
@all-contributors add @wyhaya for code

PR #525, opened 2025-06-07
Switch to RustCrypto's ChaCha20Poly1305 implementation

EpicEric:


Eugeny:
Packet sizes are encrypted: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.chacha20poly1305

EpicEric:
Thanks for pointing it out! Closing as there doesn't seem to be a way to use RustCrypto's API with the specific implementation that OpenSSH mandates.

PR #526, opened 2025-06-10
added aes128-gcm@openssh.com cipher support

handewo:
here is demo
```rust
#[tokio::main]
async fn main() -> anyhow::Result<()> {
    env_logger::builder()
        .filter_level(log::LevelFilter::Info)
        .init();
    let addrs = ("127.0.0.1", 22);

    let config = Config {
        preferred: russh::Preferred {
            cipher: Cow::Owned(vec![russh::cipher::AES_128_GCM]),
            ..Default::default()
        },
        ..<_>::default()
    };
    let mut session = client::connect(Arc::new(config), addrs, Client).await?;

    let client_key = ssh_key::PrivateKey::read_openssh_file(Path::new("/home/root/.ssh/id_ed25519"))?;
    let auth_ok = session
        .authenticate_publickey(
            "root",
            russh::keys::PrivateKeyWithHashAlg::new(Arc::new(client_key), None),
        )
        .await?;
    if auth_ok == client::AuthResult::Success {
        info!("success");
    }

    let channel = session.channel_open_session().await?;
    channel.exec(true, "cat > /dev/null").await?;

    let (mut read_half, write_half) = channel.split();
    let data = vec![45_u8; 1073741824];
    tokio::spawn(async move {
        match write_half.data(data.as_slice()).await {
            Ok(_) => {
                info!("write finish");
            }
            Err(e) => {
                info!("write failed: {}", e);
            }
        }
        write_half.eof().await.unwrap();
    });
    loop {
        if let Some(ChannelMsg::ExitStatus { exit_status }) = read_half.wait().await {
            info!("exit status: {}", exit_status);
            break;
        }
    }

    Ok(())
}

struct Client;

impl client::Handler for Client {
    type Error = russh::Error;

    async fn check_server_key(
        &mut self,
        _server_public_key: &ssh_key::PublicKey,
        Ok(true)
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ This is a fork of [Thrussh](https://nest.pijul.com/pijul/thrussh) by Pierre-Éti
 * `forward-streamlocal` (remote UNIX socket forwarding) ✨
 * Ciphers:
   * `chacha20-poly1305@openssh.com`
+  * `aes128-gcm@openssh.com` ✨
   * `aes256-gcm@openssh.com` ✨
   * `aes256-ctr` ✨
   * `aes192-ctr` ✨
--- a/russh/src/cipher/gcm.rs
+++ b/russh/src/cipher/gcm.rs
@@ -17,27 +17,26 @@
 
 use std::convert::TryInto;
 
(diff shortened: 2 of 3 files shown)


Eugeny:
Thank you!

PR #527, opened 2025-06-12
Support sending "no-more-sessions@openssh.com" requests

handewo:
This is the default behavior when openssh client connect to openssh server. 
It improves the security as openssh described in PROTOCOL:
```
2.2. connection: disallow additional sessions extension
     "no-more-sessions@openssh.com"

Most SSH connections will only ever request a single session, but a
attacker may abuse a running ssh client to surreptitiously open
additional sessions under their control. OpenSSH provides a global
request "no-more-sessions@openssh.com" to mitigate this attack.

When an OpenSSH client expects that it will never open another session
(i.e. it has been started with connection multiplexing disabled), it
will send the following global request:

       byte            SSH_MSG_GLOBAL_REQUEST
       string          "no-more-sessions@openssh.com"
       char            want-reply

On receipt of such a message, an OpenSSH server will refuse to open
future channels of type "session" and instead immediately abort the
connection.

Note that this is not a general defence against compromised clients
(that is impossible), but it thwarts a simple attack.
```
here is minimal demo:
```rust
use russh::client;
use russh::client::Config;
use russh::keys::ssh_key;
use russh::ChannelMsg;

use log::info;
use std::sync::Arc;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    env_logger::builder()
        .filter_level(log::LevelFilter::Debug)
        .init();
    let addrs = ("192.168.1.1", 22);

    let config = Config::default();
    let mut session = client::connect(Arc::new(config), addrs, Client).await?;

    let auth_ok = session.authenticate_password("root", "password").await?;
    if auth_ok != client::AuthResult::Success {
        anyhow::bail!("login failed")
    }

    let mut channel = session.channel_open_session().await?;
    channel.exec(true, "id").await?;
    session.no_more_sessions(true).await?;
    loop {
        if let Some(ChannelMsg::ExitStatus { exit_status }) = channel.wait().await {
            info!("exit status: {}", exit_status);
            break;
        }
    }
   // open session again
    let _channel = session.channel_open_session().await?;

    Ok(())
}

struct Client;

// #[async_trait]
impl client::Handler for Client {
    type Error = russh::Error;

    async fn check_server_key(
        &mut self,
        _server_public_key: &ssh_key::PublicKey,
        Ok(true)
    }
}
```
Since the demo open session twice after send '"no-more-sessions@openssh.com" request. the connection is aborted by openssh server.
here is debug log:
```
...
[2025-06-12T05:44:45Z DEBUG russh::client] disconnected: ReceivedDisconnect(RemoteDisconnectInfo { reason_code: ProtocolError, message: "Possible attack: attempt to open a session after additional sessions disabled", lang_tag: "" })
...
```

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -731,6 +731,9 @@ impl Session {
                     Some(GlobalRequestResponse::Keepalive) => {
                         // ignore keepalives
                     }
+                    Some(GlobalRequestResponse::NoMoreSessions) => {
+                        debug!("no-more-sessions@openssh.com requests success");
+                    }
                     Some(GlobalRequestResponse::TcpIpForward(return_channel)) => {
                         let result = if r.is_empty() {
                             // If a specific port was requested, the reply has no data
@@ -766,6 +769,9 @@ impl Session {
                 match self.open_global_requests.pop_front() {
                     Some(GlobalRequestResponse::Keepalive) => {
                         // ignore keepalives
+                    }
+                    Some(GlobalRequestResponse::NoMoreSessions) => {
+                        warn!("no-more-sessions@openssh.com requests failure");
                     }
                     Some(GlobalRequestResponse::TcpIpForward(return_channel)) => {
                         let _ = return_channel.send(None);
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you!

Issue #532, opened 2025-06-26
Implement faster ciphers


EpicEric:
By profiling [Sandhole](https://github.com/EpicEric/sandhole), I've figured out that the main bottleneck in terms of performance was the current implementation of ciphers. I've played around with both ChaCha20Poly1305 (the default; flamegraph below) and AES-GCM, and while the latter offered some marginal performance gains, it still felt too slow, especially when comparing Sandhole with similar implementations.

![Image](https://github.com/user-attachments/assets/dc3942cb-6340-485a-b992-25414dda9ccc)

From my tests, using an implementation of the ChaCha20Poly1305-OpenSSH cipher that [makes better use of parallelism](https://github.com/RustCrypto/traits/issues/444), we can expect Russh to have substantial speedups, up to an order of magnitude(!).

PR #533, opened 2025-06-26
Use `ring` implementation for ChaCha20Poly1305 cipher

EpicEric:
Followup to #532.

By switching from the RustCrypto implementation of ChaCha20Poly1305-OpenSSH to `ring`, my port forwarding benchmarks resulted in a substantial performance gain of an order of magnitude.

As a disclaimer, I have not tested this against an OpenSSH implementation to make sure that it's fully compatible.

---

With `russh = "0.52.1"`:

image
= 7s 933ms

---

With this PR:

image
= 662 ms

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,5 +105,5 @@ jobs:
 
     - name: Check with minimal dependency versions
       run: |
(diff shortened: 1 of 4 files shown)


EpicEric:
Superseded by #535

PR #534, opened 2025-06-26
Use `ring` implementation for AES-GCM ciphers

EpicEric:
Followup to #532. This implementation changes the signature for the `OpeningKey::open` method to not split the tag from the cipher prematurely.

As a disclaimer, I have not tested this against an OpenSSH implementation to make sure that it's fully compatible.

---

AES_256_GCM with `russh = "0.52.1"`:

image
= 6s 248ms

---

AES_256_GCM with this PR:

image
= 552ms


EpicEric:
Superseded by #535

PR #535, opened 2025-06-26
Use `aws-lc-rs`/`ring` implementation for AEAD ciphers

### LINKED ISSUE #532: Implement faster ciphers

EpicEric:
Closes #532.

By switching the ChaCha20Poly1305, AES-GCM-128 and AES-GCM-256 ciphers from the RustCrypto implementation to `ring`'s, [my port forwarding benchmarks](https://github.com/EpicEric/sandhole/blob/main/tests/https_single_stream_download.rs) resulted in a substantial performance gain of an order of magnitude on my machine.

||With current `main` | With this PR |
|-|-|-|
|ChaCha20Poly1305|8s 18ms|662 ms|
|AES-GCM-128|6s 280ms|557ms|
|AES-GCM-256|6s 396ms|552ms|

But results seem to vary depending on your hardware, where we see much smaller or no gains. On a different machine with limited memory and less CPU power (and a different benchmark):

||With current `main` | With this PR |
|-|-|-|
|ChaCha20Poly1305|660ms|654ms|
|AES-GCM-256|803ms|615ms|

This implementation also changes the signature for the `OpeningKey::open` method to not split the tag from the cipher prematurely.

I have tested ChaCha20Poly1305 and AES-GCM-256 against the current release of russh and OpenSSH, but not AES-GCM-128, although I expect it to work the same.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -571,6 +571,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "wyhaya",
+      "name": "wyhaya",
+      "avatar_url": "https://avatars.githubusercontent.com/u/23690145?v=4",
+      "profile": "https://github.com/wyhaya",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,5 +105,5 @@ jobs:
 
     - name: Check with minimal dependency versions
       run: |
-        rustup toolchain add 1.77.0
-        cargo +1.77.0 minimal-versions check --all-features --no-dev-deps
+        rustup toolchain add 1.75.0
+        cargo +1.75.0 minimal-versions check --all-features --no-dev-deps
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,5 +105,5 @@ jobs:
 
     - name: Check with minimal dependency versions
       run: |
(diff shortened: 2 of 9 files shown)


Eugeny:
Thanks! I've added direct benchmarks for the ciphers and the GCM gains are just as dramatic on my side. 

Out of curiousity, why did you decide to switch to ring from aws-lc? It seems that the general ecosystem trend is heading towards aws-lc (see rustls), especially given how it is FIPS certified.

EpicEric:
We could switch to the aws-lc-rs implementations instead, I just wasn't able to figure out how to make it compile for Windows or WASM, so I went with ring for this PR.

EpicEric:
Given that WebAssembly doesn't seem to be a supported platform for aws-lc-rs, I've replaced the dependency with ring for the target.

Perhaps choosing between either package should be a feature flag in the same vein as rustls? EDIT: I've done just that.

Eugeny:
Looking great now :+1: :+1:

PR #543, opened 2025-07-22
Fix off-by-one error for keepalive timer

EpicEric:
When doing some tests, I've noticed that using a keepalive interval of 15 seconds and a count of 3 resulted in a total of 75 seconds for the connection to be closed, instead of the expected 60 seconds. This fixes the issue by performing the addition to the internal counter before the check.


Eugeny:
Good catch, thank you!

PR #545, opened 2025-07-22
fix: allow `Stream` to connect to multiple addresses

plaflamme:
Previously, when using `config.stream()`, it would resolve the host, port pair and pick the first address returned. But Tokio has built-in support to resolve multiple addresses and connect to each one until one succeeds. So this PR lifts the previous restriction.

The main consquence is that the `Error::NotResolvable` is gone since it will be reported as an `Io` error by Tokio instead.

### DIFF
--- a/russh-config/src/lib.rs
+++ b/russh-config/src/lib.rs
@@ -5,7 +5,6 @@
     clippy::panic
 )]
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @plaflamme for code

PR #550, opened 2025-08-01
feat: add a feature toggle for `rsa`

plaflamme:
The `rsa` crate has an [open CVE](https://www.cvedetails.com/cve/CVE-2023-49092/). which seems [non-trivial to fix](https://github.com/RustCrypto/RSA/issues/19). There has been some recent progress, but it's still unclear if it will actually fix the issue.

I'm actively working on getting native SSH support in gitoxide [here](https://github.com/GitoxideLabs/gitoxide/pull/2081) and this particular CVE may become an issue for that integration.

This commit adds a feature toggle to enable/disable the `rsa` dependency.

This also fixes a race condition in the `test_agent` test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,15 +15,12 @@ home = "0.5"
 hmac = "0.12"
 log = "0.4.11"
 rand = "0.8"
-rsa = "0.9"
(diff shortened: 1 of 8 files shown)


Eugeny:
Thank you!

PR #435, opened 2024-12-30
Add an option to enable TCP_NODELAY

Patryk27:
This commit adds a new configuration option, `nodelay` - activating this option disables Nagle's algorithm for client sockets.

## Motivation

While working on [a game of mine](https://github.com/Patryk27/kartoffels) I realized that the version deployed online seems to have a noticeable lag - it didn't took long to piece two and two, and realize that Nagle's at fault. I have confirmed that calling `set_nodelay()` solves the issue on my side.

PR #551, opened 2025-08-01
Add TCP_NODELAY option for client

tom-90:
This PR adds a `nodelay` option to the russh client implementation, similar to a previous PR #435 that did this for the server implementation.

While using Warpgate as a jumphost between my DB server and my local system, I noticed a large performance impact when executing SQL queries sequentially due to high latency. Warpgate was about twice as slow in my test workload, compared to SSH directly to the server. After testing with `russh` directly, I found that by enabling this option, I could get on par performance with normal SSH.

After the next release of russh happens, I'd like to contribute some changes to Warpgate itself as well, to allow enabling `nodelay` for it's SSH server- and client-side via the config.

### DIFF
--- a/russh/examples/client_open_direct_tcpip.rs
+++ b/russh/examples/client_open_direct_tcpip.rs
@@ -10,6 +10,7 @@ use anyhow::Result;
 use clap::Parser;
 use key::PrivateKeyWithHashAlg;
 use log::info;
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @tom-90 for code

PR #557, opened 2025-08-20
Add a hook-trigger to the handler.

handewo:
## Note: this is a breaking API change.
**New feather:**
    This is useful if you need to update the handler
    without relying on a mutex (or similar synchronization).

here is demo
```rust
use tokio::sync::mpsc::Receiver;
use russh::server::Handler;

struct App{
    foo: String
    recv: Receiver<String>,
    trigger: Receiver<String>,
}

impl App for Handler {
    type Error = russh::Error;
    type Data = String;
    async fn trigger(&mut self) -> Result<Self::Data, Self::Error> {
        match self.trigger.recv().await {
            Some(d) => Ok(d),
            None => std::future::pending().await,
        }
    }

    async fn process(&mut self, s: Self::Data) -> Result<(), Self::Error> {
        let s = self.recv.recv().await.unwrap();
        self.foo = s;
        Ok(())
    }
}

```

### DIFF
--- a/russh/examples/echoserver.rs
+++ b/russh/examples/echoserver.rs
@@ -76,6 +76,7 @@ impl server::Server for Server {
 
 impl server::Handler for Server {
     type Error = russh::Error;
+    type Data = ();
 
     async fn channel_open_session(
         &mut self,
--- a/russh/examples/echoserver_certificates.rs
+++ b/russh/examples/echoserver_certificates.rs
@@ -109,6 +109,7 @@ impl server::Server for Server {
 
 impl server::Handler for Server {
     type Error = russh::Error;
+    type Data = ();
 
     async fn channel_open_session(
         &mut self,
--- a/russh/examples/ratatui_app.rs
+++ b/russh/examples/ratatui_app.rs
@@ -143,6 +143,7 @@ impl Server for AppServer {
 
 impl Handler for AppServer {
     type Error = anyhow::Error;
+    type Data = ();
 
     async fn channel_open_session(
         &mut self,
--- a/russh/examples/ratatui_shared_app.rs
+++ b/russh/examples/ratatui_shared_app.rs
@@ -145,6 +145,7 @@ impl Server for AppServer {
 
 impl Handler for AppServer {
     type Error = anyhow::Error;
+    type Data = ();
 
     async fn channel_open_session(
         &mut self,
--- a/russh/examples/sftp_server.rs
+++ b/russh/examples/sftp_server.rs
@@ -41,6 +41,7 @@ impl SshSession {
 
 impl russh::server::Handler for SshSession {
     type Error = anyhow::Error;
+    type Data = ();
 
     async fn auth_password(&mut self, user: &str, password: &str) -> Result<Auth, Self::Error> {
         info!("credentials: {user}, {password}");
--- a/russh/examples/test.rs
+++ b/russh/examples/test.rs
@@ -47,6 +47,7 @@ impl server::Server for Server {
 
 impl server::Handler for Server {
     type Error = anyhow::Error;
+    type Data = ();
 
     async fn channel_open_session(
         &mut self,
--- a/russh/src/client/test.rs
+++ b/russh/src/client/test.rs
@@ -33,6 +33,7 @@ mod tests {
 
     impl ServerHandler for TestServer {
         type Error = Error;
+        type Data = ();
 
         async fn channel_open_session(
             &mut self,
--- a/russh/src/client/test.rs
+++ b/russh/src/client/test.rs
@@ -34,6 +34,7 @@ mod tests {
 
     impl ServerHandler for TestServer {
         type Error = Error;
+        type Data = ();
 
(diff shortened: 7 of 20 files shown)


Eugeny:
IMO this is a bit of a crutch that doesn't belong in the library. Is there any reason for this except running inside the russh event loop? Mutable access to the handler? TBH I'm rather inclined to hand off resources to the task that is supposed to call `trigger()` or go all the way and introduce a custom `E` event type, adding `Msg::CustomEvent<E>` and a `fn custom_event` handler fn, integrating it into the server event loop

handewo:
Yes, Mutable access to the handler. I don’t know if there’s another way to update the handler’s status without locking it.

handewo:
I think `Msg::CustomEvent<E>` is only work in situation that ssh client support sending custom Msg. Is it?

Eugeny:
No, I meant the event loop messages that are going through `server::Session::receiver`, not wire messages

handewo:
Below is some pseudo-code whose purpose is to switch apps at runtime.  
The send_app_msg argument is passed to the app, and then `send()` is executed inside a `tokio::spawn!` block, spawned by app.  
In my view, adding `Msg::CustomEvent<E>` plus a `custom_event` handler function is not sufficient for this scenario.
```rust
pub struct MyHandler {
    app: Application,
    send_app_msg: Sender<Application>,
    recv_app_msg: Receiver<Application>,
}

pub enum Application {
    Foo(Box<Foo>),
    Bar(Box<Bar>),
}

impl ru_server::Handler for MyHandler {
    async fn data(
        &mut self,
        channel: ChannelId,
        data: &[u8],
        session: &mut ru_server::Session,
    ) -> Result<(), Self::Error> {
        match self.app {
            Application::Foo(ref mut app) => app.data(channel, data, session).await,
            Application::Bar(ref mut app) => app.data(channel, data, session).await,
        }
    }

    async fn trigger(&mut self) -> Result<Self::Data, Self::Error> {
        match self.recv_app_msg.recv().await {
            Some(d) => Ok(d),
            None => std::future::pending().await,
        }
    }

    async fn process(&mut self, app: Self::Data) -> Result<(), Self::Error> {
        app.init().await?;
        self.app = app;
    }
}
```

Eugeny:
I think this could work?


```rust
enum CustomEvent {
    AppSwap(App)
}

impl ru_server::Handler for MyHandler {
    type CustomEvent = CustomEvent;

    async fn custom_event(&mut self, e: Self::CustomEvent) {
        match e {
            CustomEvent::AppSwap(app) => {
                app.init().await?;
                self.app = app
            }
        }
    }
}

..

// from another task
session_handle.emit_custom_event(CustomEvent::AppSwap(...)) // places Msg::Custom(CustomEvent::AppSwap(...)) on the event queue
```

handewo:
Got it, I will try it

handewo:
Do you mean something like this?  
Please see the code below.  
If we introduce a generic type parameter into `Msg`, the changes required would be significant.

```rust
impl Handle {
    ...

    /// Emit custom event.
    pub async fn emit_custom_event<H>(&self, event: H::CustomEvent) -> Result<(), ()>
    where
        H: Handler,
    {
        self.sender
            .send(Msg::CustomEvent(event))
            .await
            .map_err(|_| ())
    }
}

#[derive(Debug)]
pub enum Msg<H:Handler> {
    ......
    Channel(ChannelId, ChannelMsg),
    CustomEvent(H::CustomEvent),
}

    pub(crate) async fn run<H, R>(
        mut self,
        mut stream: SshRead<R>,
        mut handler: H,
    ) -> Result<(), H::Error>
    where
        H: Handler + Send + 'static,
        R: AsyncRead + AsyncWrite + Unpin + Send + 'static,
    {
        .......
                msg = self.receiver.recv(), if !self.kex.active() => {
                    match msg {

                        Some(Msg::Disconnect {reason, description, language_tag}) => {
                            self.common.disconnect(reason, &description, &language_tag)?;
                        }
                        Some(Msg::CustomEvent(event)) => {
                            handler.custom_event(event).await?;
                        }
                     }
                  }
         .....
    }

pub trait Handler: Sized {
    type Error: From<crate::Error> + Send;
    type CustomEvent: Send;
    ...
    ///
    #[allow(unused_variables)]
    fn custom_event(
        &mut self,
        event: Self::CustomEvent,
    ) -> impl Future<Output = Result<(), Self::Error>> + Send {
        async { Ok(()) }
    }
}

```

Here is compile error
```rust
error[E0107]: missing generics for enum `server::session::Msg`
   --> russh/src/server/session.rs:358:25
    |
358 |     ) -> Result<Channel<Msg>, Error> {
    |                         ^^^ expected 1 generic argument
    |
note: enum defined here, with 1 generic parameter: `H`
   --> russh/src/server/session.rs:34:10
    |
34  | pub enum Msg<H: Handler> {
    |          ^^^ -
help: add missing generic argument
    |
358 |     ) -> Result<Channel<Msg<H>>, Error> {
    |                            +++

```

PR #565, opened 2025-09-24
fix Tabby#10301 Still cannot login with pageant

vzex:
two problems:
1. pageant protocol does not support streaming, so PageantStream should receive enough data for pageant, and then forward into.
2. [SECURITY_ATTRIBUTES] should have the correct nLength
3. [SetSecurityDescriptorOwner] in version 0.58 have bug, upgrade to 0.59 is ok.


Eugeny:
Thank you!

Eugeny:
@all-contributors add @vzex for code

PR #570, opened 2025-09-29
feat: adhere to `ssh_config` more strictly

plaflamme:
`ssh_config` allows multiple `Host` patterns to match a provided hostname. When multiple blocks match, the first value of any `key` is used.

Furthermore, some host patterns can be negated, in which case, when a host matches that block is ignored entirely.

Also, `IdentityFile` is a bit different from other parameters: every instance adds a value to the list of files.

This PR adds a more strict parsing to config files into an `SshConfig` struct which can then be queried for a specific host. The configuration is merged according to the rules above.

The new types remain private to the config module, but they could eventually be made public so clients can implement more ssh2 features.


Eugeny:
Thank you!

PR #576, opened 2025-10-07
A send_ping method for measuring latency

moshevds:
The already supported send_keepalive method is useful for keeping the connection alive, but when the application wants to measure the latency, we need to wait for the response before returning.

"ping" is a rather basic networking feature, but there is a reason why so many protocols have their own ping utility (arping, ndping, hping, httping, ws-ping, quicreach, etc.). Different networks just need different diagnostic tools.

My personal usecase is that I want to do a latency check - to a host that only serves SSH - through a SOCKS proxy. I cannot send ICMP packets because there is a proxy in between the hosts. I'm sure there are many other diagnostic use cases for this. Perhaps there are also other applications that for some reason want to show the connection latency, like 3D shooters typically can do, for example.

In any case, I think this is a useful feature for russh to have, so here I am proposing it, with this PR.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -744,6 +744,9 @@ impl Session {
                     Some(GlobalRequestResponse::Keepalive) => {
                         // ignore keepalives
                     }
+                    Some(GlobalRequestResponse::Ping(return_channel)) => {
+                        let _ = return_channel.send(());
+                    }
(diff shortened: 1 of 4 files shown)


Eugeny:
LGTM, thank you

Eugeny:
@all-contributors add @moshevds for code

PR #577, opened 2025-10-08
fix: strict kex enforce seqno on initial

kruton:
During a kex, only on the initial key exchange will the seqno likely be 1. On a re-key the sequence number could be any number.

It appears the client/kex.rs was fixed at some point, but it has some dead code. It was checking if strict-kex was negotiated or if the cause is a strict rekey, but then checked that the cause was not a rekey. This is contradictory because is_strict_rekey() can never be true if !is_rekey() is true.

This fixes the bug found while using warpgate https://github.com/warp-tech/warpgate/issues/1523


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @kruton for code

PR #578, opened 2025-10-08
fix: handle empty ssh config files

plaflamme:
Fix the case where the ssh_config is empty and also other cases where we have parameters without a `Host/Match` line.


Eugeny:
Thank you!

Issue #226, opened 2024-01-17
When is `client_address` None in `russh::server::Server::new_client`? And what to do when that happens?


samuela:
I've always found [the type](https://docs.rs/russh/latest/russh/server/trait.Server.html#tymethod.new_client)
```rs
fn new_client(&mut self, client_address: Option<std::net::SocketAddr>) -> Self::Handler {
```
confusing. Specifically: When is `client_address` ever None?

At first I took this to be a peculiarity of the API, a sort of this-should-never-happen case. But after exposing my server to traffic in the wild, I've actually witnessed a crash due to `client_address` being None. So, 

**First question:** under what circumstances is `client_address` None?

Now, supposing that `client_address` is None, the question is what to do. I'd like to simply reject all requests and immediately close these sessions. I haven't found a way to express this cleanly yet.
* `new_client` does not allow returning a Result or Error-y type, so there's no chance to raise the issue at that level.
* Perhaps I could introduce structs `MyActualHandler`, `RejectEverythingHandler` and `impl russh::server::Handler` on both of them. Then I can specify `type Handler = Box<dyn russh::server::Handler>;`. But this fails due to 
```
error[E0038]: the trait `russh::server::Handler` cannot be made into an object
  --> src/bin/sshenanigans.rs:68:22
   |
68 |   type Handler = Box<dyn russh::server::Handler>;
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ `russh::server::Handler` cannot be made into an object
   |
   = note: the trait cannot be made into an object because it requires `Self: Sized`
   = note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
```
* You might reach for https://crates.io/crates/either, and then `impl russh::server::Handler for Either<MyActualHandler, RejectEverythingHandler>` (or better yet type parameterized), but that runs into [this issue](https://stackoverflow.com/questions/35160995/implement-foreign-trait-for-foreign-type).

**Second question:** What's the recommended pattern in this case? Can type definitions/constraints be relaxed in such a way as to make this more straightforward?

samuela:
My current workaround is something like
```rs
struct OptionalHandler(Option<ServerHandler>);

#[async_trait]
impl russh::server::Handler for OptionalHandler {
  type Error = <ServerHandler as russh::server::Handler>::Error;

[...]

PR #579, opened 2025-10-09
Addresses race condition from #226

Lucy-dot-dot:
#226 identified a race condition where if a client tears down the socket before our handler is scheduled to run on the CPU again, self.new_client would receive a None value as the peer address. This PR fixes this by using the value provided by tokio and passing that to the self.new_client call wrapped in Some() to keep the API from breaking.

Such behaviour is not coming from normal clients but rather misbehaving bots that scan ports with very agressive timings.

Later PR's or if required this one, could also change the API to remove the Option from new_client but this would be a breaking change and to allow the maximum number of people to benefit from this patchm this should be done in a seperate PR.

If possible I do not want to appear in the ".all-contributorsrc" file as this is such a miniscule change, that in my opinion, it does not warrant adding me there

### DIFF
--- a/russh/src/server/mod.rs
+++ b/russh/src/server/mod.rs
@@ -869,11 +869,12 @@ pub trait Server {
                     },
                     accept_result = socket.accept() => {
                         match accept_result {
-                            Ok((socket, _)) => {
+                            Ok((socket, peer_addr)) => {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @Lucy-dot-dot for code

PR #580, opened 2025-10-09
[FIX] Ignore window adjustment after channel EOF

### LINKED ISSUE #575: Connection lost when channel is abruptly closed
The culprit is this:

https://github.com/Eugeny/russh/blob/ed78d801df786e448652dc02a38d5e4ff0bfec1e/russh/src/server/encrypted.rs#L661-L666

This condition can be triggered when, for example, using remote forwarding with `ssh -R ...`, and the TCP socket is abruptly closed by the forwarded service. Despite the client being able to send and receive more messages, this error will disconnect them immediately.

EpicEric:
Closes #575.

I don't know if this is the ideal fix, compared to, say, a check on whether the channel existed briefly before the message is handled, but it appears to fix the original issue.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -542,7 +542,7 @@ impl Session {
                         new_size = channel.recipient_window_size.saturating_add(amount);
                         channel.recipient_window_size = new_size;
                     } else {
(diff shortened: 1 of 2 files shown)


Eugeny:
LGTM, thank you!

PR #582, opened 2025-10-09
[FIX] Remove `ring` feature

EpicEric:
This is an alternative to #569, proposed in [this comment](https://github.com/Eugeny/russh/pull/569#issuecomment-3383547343). This removes the `ring` flag, making the crate the default backend when the `aws-lc-rs` feature is disabled. This way, there will always be exactly one crypto backend available.

Sadly, as per https://github.com/rust-lang/cargo/issues/2980, there is no way to set mutually-exclusive features as of yet, which means that ring always gets pulled even when it's not used.


Eugeny:
See https://github.com/Eugeny/russh/pull/569#issuecomment-3387645508

PR #585, opened 2025-10-10
Add ML-KEM post-quantum hybrid key exchange support

kruton:
Implements mlkem768x25519-sha256 PQ/T hybrid key exchange as specified in https://datatracker.ietf.org/doc/draft-ietf-sshm-mlkem-hybrid-kex/ (version 03) providing quantum-resistant SSH key exchange which protects against harvest-now-decrypt-later attacks on SSH connections. The hybrid post-quantum (ML-KEM-768) and classical (X25519) cryptography ensures that the key exchange is at least as secure as the strongest individual algorithm. This scheme is IND-CPA secure and IND-CCA2 secure under random oracle assumption.

This is implemented on top of libcrux-ml-kem which is a FIPS 203 compliant library that has been formally verified and has constant-time operations.

One difference between this PQ/T hybrid kex and previously supported kex is that the shared K is encoded as a fixed length array of bytes instead of an mpint. As such, I've introduced the SharedSecret enum and moved encoding from inside compute_keys() to the caller.

The size increases by 4.3% with both unstripped and stripped release binaries (i.e., 403KB and 303KB increase respectively). Most of the size increase comes from the libcrux-ml-kem dependency. However, given that this gives russh post-quantum crypto support, it seems worth the binary size increase.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,10 +105,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: dtolnay/rust-toolchain@nightly
     - uses: taiki-e/install-action@cargo-hack
     - uses: taiki-e/install-action@cargo-minimal-versions
 
     - name: Check with minimal dependency versions
       run: |
-        rustup toolchain add 1.75.0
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! Appreciate the tests a well 🤝

PR #586, opened 2025-10-16
Update generic-array from 0.14 to 1.x

kruton:
cargo clippy suggests the upgrade to generic-array 1.x, so this change migrates from deprecated generic-array 0.14 to 1.3.3 while maintaining compatibility with RustCrypto dependencies that still use 0.14.

Relevant changes:
 - Add compat-0_14 feature to generic-array for 0.14 compatibility
 - Update ArrayLength trait bounds: <u8> not needed
 - Use digest::generic_array for cipher/mac which uses 0.14 types
 - Replace clone_from_slice() with copy_from_slice()
 - Add OutputSizeUser trait bound to enable GenericArray with M::OutputSize


Eugeny:
Thank you!

PR #595, opened 2025-11-03
Remove unneeded deps

rukai:
* base64ct is removed, as russh is now using 2024 edition, although its not clear why it was added [here](https://github.com/Eugeny/russh/pull/523) in the first place.
* once_cell is removed, we can use LazyLock instead, which was added in rust 1.80

I raised https://github.com/Eugeny/russh/pull/596 to address the CI failure


Eugeny:
Thanks!

PR #596, opened 2025-11-03
Fix clippy lints to fix CI

rukai:
This PR contains 2 changes:
* The results of `cargo clippy --fix`
* Manual additions of `#[allow(clippy::large_enum_variant)]`, since boxing those variants would break russh's public interface, I opted to just allow the lint instead. Happy to raise a ticket to investigate boxing the variants if thats something you care about.


Eugeny:
Thanks!

PR #601, opened 2025-11-26
allow creation of RunningServer from the application

glebpom:
Sometimes it’s useful to re-implement `run_on_socket` with custom logic - for example, to read a PROXY protocol header before fully establishing the connection. To support this, we need a way to construct a `RunningServer` directly.

This PR introduces a new constructor for `RunningServer`, allowing callers to create instances manually and plug in their own connection handling logic.

### DIFF
--- a/russh/src/server/mod.rs
+++ b/russh/src/server/mod.rs
@@ -48,13 +48,13 @@ use tokio::net::{TcpListener, ToSocketAddrs};
 use tokio::pin;
 use tokio::sync::{broadcast, mpsc};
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Why is this necessary at all? The `Server` trait already only exists to house the `run_on_socket` fn and if you're reimplementing it, there's no need to use the `Server` trait at all

glebpom:
Okay, thanks. It was not clear that `Server` trait is not a part of any interface

PR #604, opened 2025-11-30
feat(kex): add shared secret retrieval and kex_done callback

stevenparkerco3:
Implement the `shared_secret_bytes` method in various KEX algorithm implementations to allow retrieval of the raw shared secret. Introduce the `kex_done` callback in the client handler to capture the shared secret and KEX algorithm after key exchange completion. This feature supports protocols that derive additional keys from the SSH shared secret.

Add tests to verify the functionality of the `kex_done` callback and ensure shared secrets are correctly captured during key exchange.


Eugeny:
LGTM except the kex_algorithm parameter can be omitted since it's already in Names

PR #606, opened 2025-12-01
fix: upgrade libcrux-ml-kem since 0.0.3 was yanked

gaetschwartz:
Version [0.0.3 of libcrux-ml-kem](https://crates.io/crates/libcrux-ml-kem/0.0.3) which prevents deps resolution. This PR upgrades it to `0.0.4`.


Eugeny:
Thanks!

PR #609, opened 2025-12-10
Fix comments handling of `russh_config::parse_ssh_config`.

tayu0110:
Hello, thank you for your great work.

I found a comment-handling issue of `russh-config`. I would like to reflect the result of investigation to main stream.

The pre-fix implementation had an issue where `found_params` became `true` when comments appeared at the top of the config, causing `HostNotFound` to be returned when processing the first Host.  
The post-fix implementation avoids this issue by explicitly ignoring comments and blank lines.

I also added a test `comment_handling`.  
The config content `CONFIG` is a valid config, that has two hosts information.  
If the comment-handling is not fixed, this test should fail because the method cannot collect configurations properly.

Since this is my first pull request, please let me know if there are any issues with the format or procedure of the request.

### DIFF
--- a/russh-config/src/lib.rs
+++ b/russh-config/src/lib.rs
@@ -203,8 +203,15 @@ fn parse_ssh_config(contents: &str) -> Result<SshConfig, Error> {
     let mut config = HostConfig::default();
     let mut found_params = false;
 
-    for line in contents.lines() {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

@all-contributors please add @tayu0110 for code

Issue #417, opened 2024-12-19
How can we send custom name Global Request?


chrong14:
when we use golang, we can do this
```golang
# create a ssh connection
sshConn, chans, reqs, err := ssh.NewClientConn(conn, "", t.sshConfig)

# use sshConn to send a Global Request named "config"
_, configerr, err := sshConn.SendRequest("config", true, conf)

```
I really want to know if russh can end custom name global request as well.

Eugeny:
No - currently there is no support for non-standard global or channel requests

PR #610, opened 2025-12-12
Add server side ping

mbund:
Similar to the client, add a `send_ping` function to the server session. This re-uses the `keepalive@openssh.com` global request, but gives a tokio channel to send back to when the server receives a response.

It can be used like so:

```rs
let (tx, rx) = tokio::sync::oneshot::channel();
let now = std::time::Instant::now();
session.send_ping(tx).unwrap();
tokio::task::spawn(async move {
    rx.await.unwrap();
    let took = std::time::Instant::now() - now;
    tracing::info!(took=?took, "ping");
});
```

I then artificially added 100ms ping to localhost (each direction):

```
sudo tc qdisc add dev lo root netem delay 100ms
```

Then I connect to the local ssh server and see logs like:

```
ping took=200.734513ms
ping took=200.458655ms
ping took=200.801369ms
ping took=200.508387ms
ping took=200.680368ms
```

Related issue: #417

My use case for this is an SSH app where I want to display the current approximate ping of the client in the TUI.

Also of note implementation-wise is that the openssh client appears to always reply to these requests with `REQUEST_FAILURE`, but I added the response handling to `REQUEST_SUCCESS` as well anyways, just in case.

### DIFF
--- a/russh/src/server/encrypted.rs
+++ b/russh/src/server/encrypted.rs
@@ -1045,6 +1045,9 @@ impl Session {
                     Some(GlobalRequestResponse::Keepalive) => {
                         // ignore keepalives
                     }
+                    Some(GlobalRequestResponse::Ping(return_channel)) => {
+                        let _ = return_channel.send(());
+                    }
                     Some(GlobalRequestResponse::TcpIpForward(return_channel)) => {
                         let result = if r.is_finished() {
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @mbund for code

PR #620, opened 2025-12-31
Make Config accessor methods for user/port/host public

packetsource:
russh_config::Config struct has gone private. But it's not unreasonable for client apps using parse_home() and friends to want to inspect the results of the excellent ssh_config parsing code, eg. host1 maps to user foo@bar:2022 etc etc., so they can print progress/diagnostic messages. There are some helpful accessor functions for user/port/host, but they are not public, so this PR simply marks those methods public.

### DIFF
--- a/russh-config/src/lib.rs
+++ b/russh-config/src/lib.rs
@@ -148,7 +148,7 @@ impl Config {
         }
     }
 
-    fn user(&self) -> String {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors add @packetsource for code

PR #624, opened 2026-01-15
Pageant WM_COPYDATA protocol fix & token tightening

vzex:
1. pageant MemoryMap always fail in cargo build --release mode if pass the security_attributes argument in MemoryMap::new, I don't know why, but pass None can fix it.
2. Support ssh certificate based authentication, when use sshagent, it will scan for both cert and publickey, if is cert, it should trigger the new auth method. example is here [https://github.com/vzex/russh/blob/main/russh/examples/client_exec_sshagent.rs](https://github.com/vzex/russh/blob/main/russh/examples/client_exec_sshagent.rs)

### DIFF
--- a/pageant/src/wmmessage.rs
+++ b/pageant/src/wmmessage.rs
@@ -233,24 +233,15 @@ pub fn query_pageant_direct(cookie: String, msg: &[u8]) -> Result<Vec<u8>, Error
     let hwnd = find_pageant_window()?;
     let map_name = format!("PageantRequest{cookie}");
 
-    let user = get_current_process_user()?;
-
-    let mut sd = SECURITY_DESCRIPTOR::default();
-    let psd = PSECURITY_DESCRIPTOR(&mut sd as *mut _ as *mut _);
-
-    unsafe {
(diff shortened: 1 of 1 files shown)


rebelliongeeks:
On a non-domain Windows 11 account, the WM_COPYDATA transport fails with 0x80070519 (ERROR_UNKNOWN_REVISION) out of the security descriptor setup in `query_pageant_direct` (debug build, pageant 0.2.1). The cause looks like a lifetime bug in `get_current_process_user`: it returns `TOKEN_USER` by value, but the `Sid` field inside it is a pointer into the local buffer Vec, which is dropped when the function returns. `SetSecurityDescriptorOwner` then receives a dangling pointer, which is undefined behavior, hence the build-mode-dependent failures. So removing the security attributes, as this PR does, is the correct fix and not just a workaround: the code being removed was handing Windows a pointer to freed memory.

For reference, connecting to Pageant's own named pipe (`\\.\pipe\pageant.<user>.<hash>`) with `connect_named_pipe` works cleanly as an alternative transport, standard agent protocol.

Issue #625, opened 2026-01-20
russh fails due to dependencies update


ahassany:
Currently, russh from main and 0.56 fails due to updates in downstream dependencies. 

```
russh git:(main) cargo test
  Downloaded shell-escape v0.1.5
  Downloaded proc-macro-error-attr v1.0.4
  Downloaded lru v0.12.5
  Downloaded numtoa v0.1.0
  Downloaded clap_lex v0.2.4
  Downloaded wasm-bindgen-test-shared v0.2.108
  Downloaded tokio-fd v0.3.0
  Downloaded wasm-bindgen-test-macro v0.3.58
  Downloaded unicode-truncate v1.1.0
  Downloaded whoami v1.6.1
  Downloaded wasm-bindgen-test v0.3.58
  Downloaded textwrap v0.16.2
  Downloaded flurry v0.5.2
  Downloaded russh-sftp v2.1.1
  Downloaded humantime v1.3.0
  Downloaded clap_derive v3.2.25
  Downloaded cassowary v0.3.0
  Downloaded termion v2.0.3
  Downloaded seize v0.3.3
  Downloaded clap v3.2.25
  Downloaded proc-macro-error v1.0.4
  Downloaded bstr v1.12.1
  Downloaded unicode-width v0.2.0
  Downloaded os_str_bytes v6.6.1
  Downloaded instability v0.3.11
  Downloaded indoc v2.0.7
  Downloaded globset v0.4.18
  Downloaded env_logger v0.6.2
  Downloaded signal-hook-mio v0.2.5
  Downloaded num_cpus v1.17.0
  Downloaded ratatui v0.29.0
  Downloaded 31 crates (1.9MiB) in 2.34s
   Compiling unicode-ident v1.0.22
   Compiling proc-macro2 v1.0.105
   Compiling quote v1.0.43
   Compiling libc v0.2.180
   Compiling version_check v0.9.5
   Compiling cfg-if v1.0.4
   Compiling typenum v1.19.0
   Compiling subtle v2.6.1
   Compiling const-oid v0.9.6
   Compiling base64ct v1.8.3
   Compiling log v0.4.29
   Compiling autocfg v1.5.0
   Compiling serde_core v1.0.228
   Compiling pin-project-lite v0.2.16
   Compiling bytes v1.11.0
   Compiling memchr v2.7.6
   Compiling rustversion v1.0.22
   Compiling libm v0.2.15
   Compiling futures-sink v0.3.31
[...]

PR #627, opened 2026-01-23
Improve CryptoVec performance

EpicEric:
Currently, `CryptoVec::drop` is the biggest bottleneck in performance for Sandhole. This PR adds a new private `zeroize` method that uses `std::ptr::write_bytes` under the hood, plus another private `optimization_barrier` method (based on https://github.com/RustCrypto/utils/pull/1252) that ensures that reads won't happen before writes - practically eliminating `CryptoVec::drop` from the flamegraph.

For the sake of transparency, there's some ongoing discussion about dropping `volatile_write` in the zeroize crate (https://github.com/RustCrypto/utils/issues/1269), and whether `optimization_barrier` is enough.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ futures = "0.3"
 home = "0.5"
 hmac = "0.12"
 log = "0.4.11"
-rand = "0.8"
+rand = { version = "0.9", features = ["thread_rng"] }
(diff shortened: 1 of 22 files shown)


EpicEric:
CI seems to fail due to #625

Eugeny:
Thanks - LGTM for now

PR #638, opened 2026-02-14
add ferrissh to readme as an adopter!

joshbenz:
ferrissh is an async SSH CLI scraper library for network device automation in Rust. Think like [scrapli](https://github.com/carlmontanari/scrapli) and [netmiko](https://github.com/ktbyers/netmiko). It's pretty early, but thought it would be fun to add. I assume I go in the adopters section.


Eugeny:
Thanks!

PR #639, opened 2026-02-15
fix(deps): bump rand libraries to 0.10

louib:
Version `0.10` is out and the exact pin is making it difficult to align the versions in a dependency tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ futures = "0.3"
 home = "0.5"
 hmac = "0.12"
 log = "0.4.11"
-rand = { version = "0.9", features = ["thread_rng"] }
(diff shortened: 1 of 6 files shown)


louib:
I don't think it will be possible to bump only `rand` and `rand_core` without bumping a bunch of crypto crates to release candidates.

Eugeny:
I'm finished bumping things in `ssh-key` but don't have time to finish updating this crate itself - pushed my work so far to the `bump-rand-2` branch

jgrund:
I have a PR where I carry this work forward a bit more:

https://github.com/Eugeny/russh/pull/673

I had to use a patch dep for hmac till that's resolved, but everything else compiles.

PR #642, opened 2026-02-15
fix: do not send keepalive before authentication

zeroleo12345:
fix https://github.com/Eugeny/russh/issues/640


Eugeny:
@all-contributors please add @zeroleo12345 for code

PR #646, opened 2026-02-27
Fix zlib vs zlib@openssh.com compression timing (#564)

### LINKED ISSUE #564: "zlib" compression incorrectly treated the same as "zlib@openssh.com"
### User-visible problem

If I compile and run the `echoserver` example program, and then attempt to connect to it using PuTTY with compression enabled, the connection hangs during setup. PuTTY's packet log shows that it sent its first encrypted packet, `SSH_MSG_SERVICE_REQUEST("ssh-userauth")`, and never received `SSH_MSG_SERVICE_ACCEPT` in reply.

I observed this when building from git commit be5c22e1da94d0ce94e16604f80fd07c0c0e8fe9, the location of the `v0.54.3` tag.

### Background

There are two related compression methods (other than `none`) supported by both PuTTY and russh. They are `zlib`, the original one defined in RFC 4253, and `zlib@openssh.com`, a modified version invented by OpenSSH.

They are not the same. In the original `zlib`, compression is initialised at the same time as encryption, in both directions: the first packet sent after `SSH_MSG_NEWKEYS` is the first compressed packet. But `zlib@openssh.com` delays enabling compression in the client-to-server direction until userauth is complete, so that unauthenticated users can't hammer on the server's zlib decoder looking for exploits.

OpenSSH's client uses `zlib@openssh.com` in preference to `zlib`. PuTTY puts them the other way round. The client's preference order determines which method is selected, so PuTTY ends up negotiating `zlib` compression while the OpenSSH client negotiates `zlib@openssh.com`.

### Diagnosis

It looks to me as if russh is treating `zlib` as an alias for `zlib@openssh.com`, which is incorrect. When PuTTY and russh negotiate the `zlib` mode for client-to-server compression, PuTTY sends its `SSH_MSG_SERVICE_REQUEST` in compressed form, because `zlib` specifies that compression is enabled instantly at the same time as encryption. But recompiling russh with extra diagnostics, I see that when russh receives that encrypted packet, it calls `Decompress::decompress()` on a `Decompress` object which at that point still contains `Decompress::None`. So it misinterprets the `SSH_MSG_SERVICE_REQUEST` as a packet with the nonsense type 120, ignores it completely, and continues waiting for what _it_ thinks is a `SSH_MSG_SERVICE_REQUEST`.

If I hack the source code of PuTTY so that it _also_ treats `zlib` as specifying the same delayed compression as `zlib@openssh.com`, then the connection is set up successfully.

### Possible root cause

I also see in `compression.rs` that the internal identifiers for the two compression methods are confusingly the wrong way round, because the identifier containing the word `LEGACY` refers to the newer `zlib@openssh.com` and not the older `zlib`:
```rust
#[cfg(feature = "flate2")]
pub const ZLIB: Name = Name("zlib");
#[cfg(feature = "flate2")]
pub const ZLIB_LEGACY: Name = Name("zlib@openssh.com");
```

This makes me wonder if the author of this part of the code might have misinterpreted the existence of the two names as meaning that the domain-qualified name came first, and was later standardised under a bare name. That is a common pattern in other cases (particularly new crypto), but in this case, it's not what happened. `zlib` came first, and `zlib@openssh.com` is not an alias for it but a _modification_ of it.

### Fix

I haven't tried to prepare a patch against the russh code, but the shape of the fix is that if `zlib` is selected, then `init_decompress` in the server should be called at the same time as encryption is set up. Only `zlib@openssh.com` should defer it until after userauth.

If russh is unwilling to take the risk of unauthenticated users being able to try to exploit the server's zlib decoder, then an alternative fix is to remove `zlib` from the option list completely, leaving only `zlib@openssh.com`.

gui-wf:
## Summary

Fixes #564.

The `Compression` enum previously mapped both `zlib` (RFC 4253) and `zlib@openssh.com` to the same `Zlib` variant, losing the critical timing distinction between the two algorithms:

- **`zlib`** (RFC 4253) - compression activates immediately after `SSH_MSG_NEWKEYS` (initial key exchange)
- **`zlib@openssh.com`** - compression is deferred until after user authentication succeeds

Because both were treated identically, russh only activated compression after authentication, which broke interoperability with clients/servers that negotiate plain `zlib` and expect compression to start right after key exchange. This was originally reported by Simon Tatham (PuTTY maintainer).

## What this fix does

- Adds a new `Compression::ZlibOpenSSH` variant to distinguish `zlib@openssh.com` from plain `zlib`
- Adds an `is_deferred()` method that returns `true` only for `ZlibOpenSSH`
- Guards all post-authentication compression activation points with `is_deferred()` checks, so they only fire for the deferred variant
- Activates non-deferred (`zlib`) compression immediately in `newkeys_received()`, right after key exchange completes

### Files changed

- `russh/src/compression.rs` - new `ZlibOpenSSH` variant, updated `new()` constructor, `is_deferred()` method
- `russh/src/session.rs` - activate non-deferred compression at key exchange time
- `russh/src/client/encrypted.rs` - guard deferred decompression init with `is_deferred()`
- `russh/src/client/mod.rs` - guard deferred compression init with `is_deferred()`
- `russh/src/server/encrypted.rs` - guard deferred compress/decompress init with `is_deferred()`

## Test plan

- All existing tests pass, including the compression integration test (`cargo test`)
- The fix is a minimal, targeted change that preserves existing `zlib@openssh.com` behavior while correctly adding the RFC 4253 `zlib` immediate-activation path

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -109,7 +109,9 @@ impl Session {
                                 .send(Reply::AuthSuccess)
                                 .map_err(|_| crate::Error::SendError)?;
                             enc.state = EncryptedState::InitCompression;
-                            enc.server_compression.init_decompress(&mut enc.decompress);
+                            if enc.server_compression.is_deferred() {
+                                enc.server_compression.init_decompress(&mut enc.decompress);
+                            }
                             return Ok(());
                         }
                         Some((&msg::USERAUTH_BANNER, mut r)) => {
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -1188,8 +1188,10 @@ impl Session {
 
             if let Some(ref mut enc) = self.common.encrypted {
                 if let EncryptedState::InitCompression = enc.state {
-                    enc.client_compression
-                        .init_compress(self.common.packet_writer.compress());
+                    if enc.client_compression.is_deferred() {
+                        enc.client_compression
+                            .init_compress(self.common.packet_writer.compress());
+                    }
                     enc.state = EncryptedState::Authenticated;
                 }
             }
--- a/russh/src/compression.rs
+++ b/russh/src/compression.rs
@@ -8,6 +8,8 @@ pub enum Compression {
     None,
     #[cfg(feature = "flate2")]
     Zlib,
+    #[cfg(feature = "flate2")]
+    ZlibOpenSSH,
 }
 
 #[derive(Debug)]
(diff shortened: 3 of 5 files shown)


Eugeny:
Thank you! 

@all-contributors please add @gui-wf for code

PR #652, opened 2026-03-02
Expose HostConfig fields to external consumers

fbernier:
I need `ProxyJump` information for something I work on, and this is not available in `russh-config`. This crate is quite useful for parsing an ssh config without shelling out to `ssh -W` but as it is now it completely hides away most of the parsed output.

I don't see a good reason for not making everything available to I made everything public. Let me know if I'm missing anything.

### DIFF
--- a/russh-config/src/lib.rs
+++ b/russh-config/src/lib.rs
@@ -26,25 +26,25 @@ mod proxy;
 pub use proxy::*;
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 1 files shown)


fbernier:
just noticed that this would fix https://github.com/Eugeny/russh/issues/593.

Eugeny:
Thanks! @all-contributors please add @fbernier for code

PR #653, opened 2026-03-03
perf: eliminate mlock/munlock overhead for non-secret buffers (~21% throughput improvement)

### LINKED ISSUE #629: RFC: CryptoVec mlock behavior - should it be secure by default or opt-in?
CryptoVec currently mlocks all allocated memory to prevent swapping sensitive data to disk. While secure, this has significant performance overhead for high-throughput use cases like SFTP file transfers.

  Profiling shows mlock/munlock syscalls using a lot of CPU time in one of my projects (25%+) when most of the data flowing through buffers isn't actually cryptographic secrets - it's packet payloads, channel data (stdin/stdout), etc.

  ## OpenSSH's approach

  OpenSSH's `sshbuf` uses non-mlocked buffers by default:
  - `sshbuf_new()` at [packet.c:245-248](https://github.com/openssh/openssh-portable/blob/master/packet.c#L245-L248) creates
  buffers for `incoming_packet`, `outgoing_packet`, etc. **without** mlock
  - Channel data at [channels.c:3479](https://github.com/openssh/openssh-portable/blob/master/channels.c#L3479) uses regular
  buffers
  - Sensitive data (private keys, shared secrets) gets explicit secure handling via `freezero()`
  ([misc.c](https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/freezero.c))

  ## Proposed changes

  ### 1. Consistent zeroization

  The `resize()` truncation path uses [`memset` at cryptovec.rs:216](https://github.com/Eugeny/russh/blob/main/cryptovec/src/cryptovec.rs#L216) which could be optimized away by the compiler. However, reallocation uses [`zeroize` at line 230](https://github.com/Eugeny/russh/blob/main/cryptovec/src/cryptovec.rs#L230) and Drop uses [`zeroize` at line 367](https://github.com/Eugeny/russh/blob/main/cryptovec/src/cryptovec.rs#L367), both with the [optimization barrier at line 383](https://github.com/Eugeny/russh/blob/main/cryptovec/src/cryptovec.rs#L383).

  Should use `zeroize()` consistently in all paths, matching:
  - OpenSSH's [explicit_bzero](https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/explicit_bzero.c) pattern
  - RustCrypto's [optimization_barrier approach](https://github.com/RustCrypto/utils/pull/1252) (see also [issue
  #1269](https://github.com/RustCrypto/utils/issues/1269))

  ### 2. Conditional mlock with categorized buffers

  Add a `secure: bool` field to CryptoVec and categorize buffers:

   ### Should be mlocked (secrets)

  - **Shared secrets (K)** - passed to `kex_derive_keys()`, cleared with `explicit_bzero` after use
  ([kex.c:1123-1145](https://github.com/openssh/openssh-portable/blob/master/kex.c#L1123-L1145))
  - **Derived session keys** - `newkeys->enc.key`, `newkeys->mac.key` cleared with `explicit_bzero`, struct freed with `freezero`   ([kex.c:690-711](https://github.com/openssh/openssh-portable/blob/master/kex.c#L690-L711))
  - **Exchange hash working buffer** - contains K during derivation via `ssh_digest_update_buffer(hashctx, shared_secret)`
  ([kex.c:1078-1097](https://github.com/openssh/openssh-portable/blob/master/kex.c#L1078-L1097))
  - **Agent lock password** - `lock_pwhash` cleared with `explicit_bzero` after use
  ([ssh-agent.c:1474-1485](https://github.com/openssh/openssh-portable/blob/master/ssh-agent.c#L1474-L1485))

  ### Should NOT be mlocked (not secrets)

  - **Packet I/O buffers** - `incoming_packet`, `outgoing_packet` created with `sshbuf_new()`, no mlock
  ([packet.c:245-248](https://github.com/openssh/openssh-portable/blob/master/packet.c#L245-L248))
  - **Channel data** - `c->input`, `c->output`, `c->extended` created with `sshbuf_new()`, no mlock
  ([channels.c:540-542](https://github.com/openssh/openssh-portable/blob/master/channels.c#L540-L542))
  - **Decompression buffer** - `compression_buffer` created with `sshbuf_new()`, no mlock
  ([packet.c:800-801](https://github.com/openssh/openssh-portable/blob/master/packet.c#L800-L801))
  - **Public keys** - regular `sshbuf` allocation, `sshbuf_free` uses `freezero` for zeroization but no mlock
  ([sshbuf.c:191-192](https://github.com/openssh/openssh-portable/blob/master/sshbuf.c#L191-L192))

  Note: OpenSSH's `sshbuf_new()` ([sshbuf.c:93](https://github.com/openssh/openssh-portable/blob/master/sshbuf.c#L93)) does not
  mlock, but `sshbuf_free()` ([sshbuf.c:191](https://github.com/openssh/openssh-portable/blob/master/sshbuf.c#L191)) uses
  `freezero()` to zero memory. The model is: zero everything on free, but only mlock actual secrets.

  ## Question: What should the API default be?

  **Option A: Secure by default**

  ```rust
  CryptoVec::new()                  // mlocked
  CryptoVec::new_unlocked()         // not mlocked, for non-secrets
  CryptoVec::from_slice_unlocked()  // not mlocked
  CryptoVec::from_vec_unlocked()    // not mlocked
```
  - Safer for library consumers who may not think about security
  - External code using CryptoVec would need to opt-out for performance

  Option B: Performance by default (matches OpenSSH)
```rust
  CryptoVec::new()           // not mlocked (like sshbuf_new())
  CryptoVec::new_secure()    // mlocked, for secrets
  CryptoVec::from_slice_secure()
  CryptoVec::from_vec_secure()
```
  - Matches OpenSSH's model
  - External code would need to opt-in for security

Which approach do you prefer? I'm happy to implement either way.

mjc:
## Motivation

Every `CryptoVec` allocation calls `mlock(2)` to pin the memory and every
deallocation calls `munlock(2)` + `explicit_bzero`. For secret data (private
keys, shared secrets, derived session keys) that cost is justified. For the
high-frequency data path — incoming channel payloads, packet framing buffers,
compression scratch space, authentication algorithm names — it is pure overhead.

### `CryptoVec` → `Vec<u8>` for non-secret buffers

Applies OpenSSH's own rule: only secret material needs memory protection.
OpenSSH allocates all general packet buffers with plain `sshbuf_new()` (no
`mlock`) — see [packet.c L241-244](https://github.com/openssh/openssh-portable/blob/bb781f0/packet.c#L241).

Non-secret buffers changed: `SSHBuffer`, `PacketWriter::write_buffer` and
`compress_buffer`, `IncomingSshPacket`, `Exchange` kex public values,
`SIGNATURE_BUFFER` (signature verification scratch), `pending_reads`,
`CurrentRequest` key/algo fields, `OpenChannelMessage` helpers, all ephemeral
DH public keys.

Secret material that stays `CryptoVec` (mlocked): `session_id`,
`SharedSecret`, key derivation scratch buffers (`KEY_BUF`, `NONCE_BUF`,
`MAC_BUF`, `BUFFER`), agent `Lock` passphrase.

### `CryptoVec` → `Bytes` for `ChannelMsg::Data`

Channel payloads are application data, not key material. Using `bytes::Bytes`
eliminates the allocation + mlock on the receive path **and** allows payloads to
be forwarded zero-copy downstream.

OpenSSH never mlocks channel data — it reads via `sshpkt_get_string_direct` (a
pointer into the already-decrypted packet buffer, no mlock) — see
[channels.c L3476-3494](https://github.com/openssh/openssh-portable/blob/bb781f0/channels.c#L3476).

`Session::data` / `Session::extended_data` now accept `impl Into<bytes::Bytes>`,
so callers that already hold a `Bytes` incur no copy.

### `zeroize` for secure memory clearing in `CryptoVec`

Replaced manual `ptr::write_volatile` loop with `zeroize::Zeroize`, which uses
compiler barriers to prevent dead-store elimination — matching OpenSSH's
`explicit_bzero` /
[freezero](https://github.com/openssh/openssh-portable/blob/bb781f0/openbsd-compat/freezero.c#L25)
pattern.

## Benchmark

SFTP roundtrip throughput (loopback, chacha20-poly1305@openssh.com, 256 MB file,
8 rounds each), both runs on an idle machine back-to-back.

**`main`:**

| Run | Upload (MB/s) | Download (MB/s) | Roundtrip (MB/s) |
|-----|--------------|-----------------|-----------------|
| r1  | 298.0        | 576.6           | 392.9           |
| r2  | 386.7        | 576.6           | 462.9           |
| r3  | 566.4        | 588.5           | 577.2           |
| r4  | 560.2        | 613.9           | 585.8           |
| r5  | 225.6        | 401.9           | 288.9           |
| r6  | 233.4        | 580.5           | 332.9           |
| r7  | 379.3        | 592.6           | 462.5           |
| r8  | 555.3        | 563.9           | 559.6           |
| **avg** | **400.6** | **561.8**   | **457.8**       |

**This PR:**

| Run | Upload (MB/s) | Download (MB/s) | Roundtrip (MB/s) |
|-----|--------------|-----------------|-----------------|
| r1  | 759.6        | 707.2           | 732.5           |
| r2  | 658.1        | 701.4           | 679.0           |
| r3  | 663.2        | 693.8           | 678.1           |
| r4  | 649.7        | 711.1           | 679.0           |
| r5  | 673.7        | 679.0           | 676.4           |
| r6  | 659.8        | 713.1           | 685.4           |
| r7  | 619.9        | 632.1           | 625.9           |
| r8  | 577.9        | 668.4           | 619.9           |
| **avg** | **657.7** | **688.3**   | **672.0**       |

**Improvement: +64% upload, +23% download, +47% roundtrip**

The mlock stalls are clearly visible in `main` upload runs (r1/r5/r6 at 225–298
MB/s) — these are `mlock`/`munlock` syscalls blocking the send path on every
packet received. `main` download is less affected because the read path was
already less chatty with CryptoVec.

After this PR, upload variance is greatly reduced (577–759 MB/s) and the
remaining bottleneck is symmetric crypto (~45% of CPU at profiling time).

## API impact

- `ChannelMsg::Data { data }` and `ChannelMsg::ExtendedData { data }`: field
  type changed `Vec<u8>` → `bytes::Bytes`. Callers can wrap with
  `Bytes::from(vec)` (zero-copy).
- `Session::data` / `Session::extended_data`: now generic over
  `impl Into<bytes::Bytes>` — accepts `Vec<u8>`, `&[u8]`, `Bytes`, or
  `CryptoVec` unchanged.
- `write_kex` return type: `CryptoVec` → `Vec<u8>`.

Closes #629

### DIFF
--- a/russh/examples/ratatui_shared_app.rs
+++ b/russh/examples/ratatui_shared_app.rs
@@ -36,7 +36,7 @@ impl TerminalHandle {
         let (sender, mut receiver) = unbounded_channel::<Vec<u8>>();
         tokio::spawn(async move {
             while let Some(data) = receiver.recv().await {
-                let result = handle.data(channel_id, data.into()).await;
+                let result = handle.data(channel_id, data).await;
                 if result.is_err() {
                     eprintln!("Failed to send data: {result:?}");
                 }
--- a/russh/src/cipher/mod.rs
+++ b/russh/src/cipher/mod.rs
@@ -233,6 +233,7 @@ pub(crate) trait SealingKey {
         buffer.buffer.extend_from_slice(payload);
         let pad_offset = buffer.buffer.len();
         buffer.buffer.resize(pad_offset + padding_length, 0);
+        #[allow(clippy::indexing_slicing)] // length checked
         self.fill_padding(&mut buffer.buffer[pad_offset..]);
         let tag_offset = buffer.buffer.len();
         buffer.buffer.resize(tag_offset + self.tag_len(), 0);


mjc:
Several reviewer comments flagged that `Session::data` now takes `impl Into<bytes::Bytes>` but `CryptoVec` doesn't implement `Into<Bytes>`. I decided **not** to add that conversion.

Channel data is application payload, not secret key material. Callers that were previously passing `CryptoVec` were only doing so because the old API required it — they were all wrapping plain data with `CryptoVec::from_slice(data)` or similar. The migration is mechanical: replace with `data` (`&[u8]` coerces), `Bytes::copy_from_slice(&cv)`, or `Bytes::from(vec)`. Adding `From<CryptoVec> for Bytes` would suggest it's normal to forward mlock'd secret buffers as channel payloads, which is the opposite of what this PR is trying to establish.

That said if we don't want to break that I'll adapt the PR.

EpicEric:
The LLM-generated comments feel too excessive and only relevant for the PR at hand, not for the codebase in the long-term

mjc:
@EpicEric yeah, I added those to make reviewing a little easier while I was working on it, happy to reduce them.

mjc:
Since other merges happened while I was busy with other work, I rebased and made those changes

EpicEric (changes requested):
The ratatui_*app.rs examples need a slight change in order to compile:

```diff
impl TerminalHandle {
    async fn start(handle: Handle, channel_id: ChannelId) -> Self {
        let (sender, mut receiver) = unbounded_channel::<Vec<u8>>();
        tokio::spawn(async move {
            while let Some(data) = receiver.recv().await {
-               let result = handle.data(channel_id, data.into()).await;
+               let result = handle.data(channel_id, data).await;
                if result.is_err() {
                    eprintln!("Failed to send data: {result:?}");
                }
            }
        });
        Self {
            sender,
            sink: Vec::new(),
        }
    }
}

Eugeny:
Thanks all for the efforts and the discussion :+1:

@all-contributors please add @mjc for code

PR #655, opened 2026-03-06
Add russh/serde feature to enable serde on russh::keys::PublicKey

kpcyrd:
This way I don't need to add a dependency on `internal-russh-forked-ssh-key` to enable the `serde` feature (and keep the version constrain in-sync) and instead can just do it through `russh` :)


Eugeny:
Thanks!

PR #656, opened 2026-03-06
Remove heap allocations from SshId

kpcyrd:
The config currently requires the `server_id` on the heap (and `Config::default()` creates an allocation just for that), with this change it's pointing to an address in the compiled binary by default.

In praxis this barely matters because it's most likely only done once, but while doing this I also noticed `write` would allocate a temporary string every time it's used, so everytime an incoming connection is accepted, or the client connects to a remote server. I've removed that allocation too, so the ssh handshake should get a tiny bit faster with this change.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -679,6 +679,24 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "Mota-Link",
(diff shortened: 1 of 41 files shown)


Eugeny:
Thanks! @all-contributors please add @kpcyrd for code

PR #660, opened 2026-03-17
Replace libcrux-ml-kem with RustCrypto ml-kem

kpcyrd:
This replaces the dependency on the libcrux crate with the [RustCrypto](https://github.com/RustCrypto/) ml-kem crate.

Motivation is that I'm packaging russh for Debian and it's easier for me to deal with RustCrypto crates instead of packaging the libcrux stack.

With this change, the `cargo debstatus` output goes from this:

<details>

image

</details>

To this:

image

I've tested it like this:
```
ssh -vvv -o KexAlgorithms=mlkem768x25519-sha256 -p 2222 127.0.0.1
```

Debug logs of the key exchange:
```
debug3: kex_choose_conf: will use strict KEX ordering
debug1: kex: algorithm: mlkem768x25519-sha256
debug1: kex: host key algorithm: ssh-ed25519
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: kex: client->server cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug3: send packet: type 30
debug1: expecting SSH2_MSG_KEX_ECDH_REPLY
debug3: receive packet: type 31
debug1: SSH2_MSG_KEX_ECDH_REPLY received
debug1: Server host key: ssh-ed25519 SHA256:N+oDuZmoOF3f/kzObT9bVtrL0qElKpmDzq9B2gJzqNQ
```

CC'ing the author of the original implementation: @kruton

There's also a [recent blogpost about libcrux](https://symbolic.software/blog/2026-02-05-cryspen/), but this wasn't the motivation for this patch, it's really only so I don't have to deal with stuff like `libcrux-sha3` when there's already `sha3` ~~at home~~ in Debian.

### DIFF
--- a/russh/Cargo.toml
+++ b/russh/Cargo.toml
@@ -89,7 +89,7 @@ yasna = { version = "0.5.0", features = [
   "num-bigint",
 ], optional = true }
 zeroize = "1.7"
-criterion = { version = "0.3", optional = true, features = ["html_reports"] }
+criterion = { version = "0.4", optional = true, features = ["html_reports"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { workspace = true, features = [


kruton:
I'm okay with the concept of replacing the libcrux version, but haven't reviewed this change.

Eugeny:
In general I have no objections and can merge it unless @kruton  notices any issues

kpcyrd:
LGTM! :+1:

PR #634, opened 2026-02-05
CryptoVec: replace memset with zeroize in resize() method

EpicEric:
Fixes #631

PR #661, opened 2026-03-17
Harden Windows memory locking: fix ERROR_WORKING_SET_QUOTA and edge cases

### LINKED ISSUE #504: Log warning on windows
I've compiled a project using russh, and used it on linux without any issue.
But then I cross-compiled it for windows, and run it, getting a warning at runtime.
The warning is "russh_cryptovec::platform::error: mor.rs:37 Security warning: OS has failed to lock/unlock memory for cryptographic buffer: VirtualLock 0x5ad".

The official windows documentation writes that this error means: 
```
ERROR_WORKING_SET_QUOTA

1453 (0x5AD)

Insufficient quota to complete the requested service.
```

Then on [this](https://www.luxford.com/general_windows#h.ev2xj2re17iu) link it is written that this error happens when:
```
Call SetProcessWorkingSet with the total from step 2 plus some additional buffer to ensure other key areas - such as the stack guard page - are included.  If this step is skipped, VirtualAlloc will fail with error code 1453 (0x5ad) or ERROR_WORKING_SET_QUOTA:  Insufficient quota to complete the requested service.
```
So i suspect that the requested space does not take in account some additional memory for those services.

I'll look it more deeply in a couple of days, but in the meantime I prefered to share the issue so that someone else could work on this problem.

coreyleavitt:
## Problem

On Windows, `VirtualLock` fails with `ERROR_WORKING_SET_QUOTA` (0x5ad) on default system configurations because the process minimum working set (~200KB) leaves no headroom for locked pages. This causes the security warning:

> Security warning: OS has failed to lock/unlock memory for a cryptographic buffer: VirtualLock: 0x5ad

While investigating this, I also noticed a few edge cases in the existing implementation that could cause issues in certain scenarios.

Reported in #504.

## Changes

All changes are in `cryptovec/src/platform/windows.rs` and `cryptovec/Cargo.toml`. The public API (`mlock`/`munlock` signatures) is unchanged.

### Working set growth on demand (fixes #504)

When `VirtualLock` fails with `ERROR_WORKING_SET_QUOTA`, `lock_page` now grows the process working set by one page via `SetProcessWorkingSetSizeEx` and retries once. This is the pattern recommended by [Microsoft's VirtualLock documentation](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtuallock):

> Applications that need to lock larger numbers of pages must first call the SetProcessWorkingSetSize function to increase their minimum and maximum working set sizes.

Working set flags from `GetProcessWorkingSetSizeEx` are masked to documented `QUOTA_LIMITS_HARDWS_*` bits before passing back to avoid forwarding undocumented bits from future Windows versions.

### Partial mlock rollback

If `mlock` is locking pages A, B, C and page C fails, the previous implementation left pages A and B locked with orphaned refcounts. Now both newly locked pages and refcount increments on already-locked pages are tracked and rolled back on failure.

### Locked page cap

Total locked pages are capped at 256 (~1 MiB on 4 KiB page systems) to prevent unbounded working set growth.

### Edge case fixes

- **Page range overflow**: `get_page_range` now uses last-byte address calculation instead of `len + page_size - 1`, which could overflow when `len` is near `usize::MAX`.
- **Address overflow**: Page address computed via `checked_mul` instead of bare `*` to prevent silent wraparound on 32-bit targets.
- **VirtualLock size**: Changed from `1` byte (relies on OS rounding up) to `page_size` for explicit page-granularity semantics.
- **SYSTEM_INFO init**: Changed from `Default` to `mem::zeroed()` for guaranteed zero-initialization.

### Dependency updates

- Replaced `winapi` 0.3 with `windows-sys` 0.59 (already in the dependency tree).
- Removed `libc` dependency (`memset` was already replaced by `zeroize` in #634).

### Tests

Added unit tests for page range arithmetic (cross-platform) and mlock/munlock integration (runs on the existing Windows CI): roundtrip locking, refcount overlap handling, zero-length no-op, multiple distinct buffers exercising working set growth, and munlock-without-mlock error path.

### Other improvements

- `log::debug!` on working set growth events.
- Module-level and per-function documentation.
- Error messages include symbolic names alongside hex codes.
- Unsafe blocks scoped to individual FFI calls.

## Relationship to #653

Builds on #653, which reduced mlock to only the buffers that hold secret material. This PR hardens those remaining mlock calls on Windows so they succeed on default system configurations and handle edge cases correctly.

Fixes #504

### DIFF
--- a/cryptovec/src/platform/windows.rs
+++ b/cryptovec/src/platform/windows.rs
@@ -8,8 +8,8 @@
 //! working set is too small), the process working set is grown incrementally
 //! and the lock is retried.
 
-use std::collections::btree_map::Entry;
 use std::collections::BTreeMap;
+use std::collections::btree_map::Entry;
 use std::sync::{Mutex, OnceLock};
 
 use windows_sys::Win32::System::Memory::{
@@ -136,9 +136,10 @@ fn lock_page(
         )));
     }
 
-    let addr = page_idx.checked_mul(page_size).ok_or_else(|| {
-        MemoryLockError::new("VirtualLock: page address overflow".into())
-    })? as *mut std::ffi::c_void;
+    let addr = page_idx
+        .checked_mul(page_size)
+        .ok_or_else(|| MemoryLockError::new("VirtualLock: page address overflow".into()))?
+        as *mut std::ffi::c_void;
 
     // First attempt.
     if unsafe { VirtualLock(addr, page_size) } != 0 {
@@ -172,9 +173,10 @@ fn lock_page(
 
 /// Unlock a single page.
 fn unlock_page(page_idx: usize, page_size: usize) -> Result<(), MemoryLockError> {
-    let addr = page_idx.checked_mul(page_size).ok_or_else(|| {
-        MemoryLockError::new("VirtualUnlock: page address overflow".into())
-    })? as *mut std::ffi::c_void;
+    let addr = page_idx
+        .checked_mul(page_size)
+        .ok_or_else(|| MemoryLockError::new("VirtualUnlock: page address overflow".into()))?
+        as *mut std::ffi::c_void;
     if unsafe { VirtualUnlock(addr, page_size) } == 0 {
         let err = unsafe { windows_sys::Win32::Foundation::GetLastError() };
         return Err(MemoryLockError::new(format!(


Eugeny:
LGTM and thank you for the clean implementation :+1:

Eugeny:
@all-contributors add @coreyleavitt for code

PR #662, opened 2026-03-17
fix: use remote channel ID in CHANNEL_REQUEST replies

### LINKED ISSUE #567: russh is not compatible with jumpserver
Hi,
Here is an issue of tabby, a popular ssh terminal : https://github.com/Eugeny/tabby/issues/10228#issuecomment-3223715794
russy works well when connecting SSH directly, but fail to run when connecting  SSH through jumpserver.

Analysis and Solutions for JumpServer SSH Connection Disconnect Issues with Tabby Client

Original post
Published on 2025-08-22 09:14:13
· 1k views
· CC 4.0 BY-SA License

Project link: [JumpServer on GitCode](https://gitcode.com/feizhiyun/jumpserver)

JumpServer — a widely popular open-source bastion host

Problem Background

Recently, users reported an abnormal disconnection issue when using the Tabby client (version 1.0.223) to connect to JumpServer via SSH. Specifically, the connection disconnects automatically just a few seconds after being established. This issue does not occur when connecting to other SSH services.

Technical Analysis

After in-depth investigation, the root cause was identified as a compatibility problem between the russh SSH client library adopted by Tabby since version 1.0.216 and JumpServer’s keepalive mechanism.

Key Findings

Client change: Tabby switched to using russh as its SSH client library starting from version 1.0.216.

Issue characteristic: The rapid disconnection issue only occurs when connecting to JumpServer, while other SSH servers work fine.

Temporary workaround: Disabling keepalive in the Tabby client avoids the disconnects.

Root cause: The russh client likely cannot properly handle keepalive response packets sent by JumpServer.

Solutions
Solution 1: Adjust JumpServer Configuration

By setting the following parameter in JumpServer’s configuration, server-side keepalive heartbeats can be disabled:

CLIENT_ALIVE_INTERVAL=0

This setting prevents JumpServer from sending keepalive requests to clients, thus avoiding the compatibility problem.

Solution 2: Adjust Client Configuration

Disable the keepalive feature in Tabby:

Open Tabby settings.

Navigate to the SSH connection configuration.

Find and disable the keepalive option.

Solution 3: Downgrade or Wait for Fix

Since the issue was introduced in a specific Tabby version, users can:

Temporarily downgrade to Tabby version 1.0.215 or earlier.

Monitor Tabby’s updates and wait for an official fix.

In-depth Technical Explanation

The SSH keepalive mechanism is mainly used to maintain connection liveliness during long periods of inactivity. Typically, either the client or server periodically sends specific heartbeat packets to check the connection status. When one side fails to properly process keepalive packets from the other side, unexpected connection drops may occur.

JumpServer, as a jump host system, may implement SSH keepalive handling differently from the standard OpenSSH, and the russh client library currently does not fully handle these subtle differences, which leads to compatibility issues.

Summary

The compatibility issue between JumpServer and the newer Tabby client versions primarily stems from differences in handling SSH keepalive packets. By adjusting keepalive settings either on the server or client side, the connection disconnect problem can be effectively resolved.

It is recommended that users choose the most appropriate solution based on their environment and continue to monitor updates from the related projects to improve their experience.

For enterprise users, it is advisable to test configuration changes in a staging environment before applying them in production, ensuring system stability and security are maintained.

Mota-Link:
Fixes #567 

CHANNEL_SUCCESS and CHANNEL_FAILURE replies to incoming CHANNEL_REQUEST messages wrote the local (client-side) channel ID instead of the remote (server-side) channel ID.

Per RFC 4254 §5.4, the "recipient channel" field must contain the channel number assigned by the sender of the original request, i.e. the server-side ID stored in `ChannelParams::recipient_channel`.

When multiple channels exist and the local and remote IDs diverge, the server cannot match the reply to any known channel. Servers with keepalive enabled (e.g. OpenSSH ClientAliveInterval) interpret this as an unresponsive client and disconnect after ClientAliveCountMax failed attempts.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -508,10 +508,12 @@ impl Session {
                             if let Some(ref mut enc) = self.common.encrypted {
                                 trace!("Received channel keep alive message: {req:?}",);
                                 self.common.wants_reply = false;
-                                push_packet!(enc.write, {
-                                    map_err!(msg::CHANNEL_SUCCESS.encode(&mut enc.write))?;
-                                    map_err!(channel_num.encode(&mut enc.write))?;
-                                });
+                                if let Some(ch) = enc.channels.get(&channel_num) {
+                                    push_packet!(enc.write, {
+                                        map_err!(msg::CHANNEL_SUCCESS.encode(&mut enc.write))?;
+                                        map_err!(ch.recipient_channel.encode(&mut enc.write))?;
+                                    });
+                                }
                             }
                         } else {
                             warn!("Received keepalive without reply request!");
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

@all-contributors please add @Mota-Link for code

PR #666, opened 2026-03-17
fix: accept full 256k channel packets

mjc:
## Summary

I ran into this preventing me from doing 1GB transfers from openssl-portable's `sftp` client to my own sftp server w/ memory backend.

Changes in this PR:

- allow SSH transport packet lengths slightly above 256 KiB so full 256 KiB channel packets plus framing/padding are accepted
- add a regression test covering full 256 KiB channel packets over `aes256-gcm@openssh.com`
- keep the change narrowly scoped to the packet-length limit

## OpenSSH reference point
- [`openssh-portable` enforces a `PACKET_MAX_SIZE (256 * 1024)` transport cap in `packet.c`](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/packet.c#L99)
- [incoming packets are rejected when `packet_length > PACKET_MAX_SIZE` in the transport read path](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/packet.c#L1650-L1692)
- OpenSSH's default session-channel limit is smaller: [`CHAN_SES_PACKET_DEFAULT = 32 * 1024` in `channels.h`](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/channels.h)
- so this PR is not copying OpenSSH's exact channel defaults; it makes `russh`'s advertised `256 KiB` channel packet size internally consistent with SSH transport framing and padding

## Minimal repro
- `russh/tests/test_max_channel_packet_size.rs` fails when copied onto `main` and run as `cargo test -p russh test_aes256_gcm_allows_full_256k_channel_packet -- --nocapture`
- the same test passes on `fix-256k-packet-limit`
- baseline failure mode: `channel closed before echoing a full 256 KiB packet`

## Validation
- `cargo test -p russh`
- targeted fail/pass comparison of `test_aes256_gcm_allows_full_256k_channel_packet` on `main` vs `fix-256k-packet-limit`

### DIFF
--- a/russh/src/cipher/mod.rs
+++ b/russh/src/cipher/mod.rs
@@ -311,10 +311,13 @@ pub(crate) async fn read<R: AsyncRead + Unpin>(
 pub(crate) const PACKET_LENGTH_LEN: usize = 4;
 
 const MINIMUM_PACKET_LEN: usize = 16;
+// Keep the transport limit aligned with the 256 KiB channel packet baseline.
 const MAXIMUM_PACKET_LEN_BASELINE: usize = 256 * 1024;
 const CHANNEL_DATA_PACKET_OVERHEAD: usize = 1 + 4 + 4;
 const CHANNEL_EXTENDED_DATA_PACKET_OVERHEAD: usize = CHANNEL_DATA_PACKET_OVERHEAD + 4;
 const PADDING_LENGTH_LEN: usize = 1;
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #667, opened 2026-03-17
Remove Home Crate Dependency

rogkne:
The home crate provides the function home_dir() to work around a bug in std::env::home_dir() on Windows (incorrectly considering the HOME env var). Since Russh uses Rust 1.85, this issue is fixed, making the home crate unnecessary.

See: https://crates.io/crates/home/0.5.12

> Note: This has been fixed in Rust 1.85 to no longer use the HOME environment variable on Windows. If you are still using this crate for the purpose of getting a home directory, you are strongly encouraged to switch to using the standard library's [home_dir](https://doc.rust-lang.org/nightly/std/env/fn.home_dir.html) instead. It is planned to have the deprecation notice removed in 1.87.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,6 @@ digest = "0.10"
 delegate = "0.13"
 env_logger = "0.6"
 futures = "0.3"
-home = "0.5"
 hmac = "0.12"
 log = "0.4.11"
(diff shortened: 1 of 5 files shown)


Eugeny:
Thank you!

PR #670, opened 2026-03-18
Fix and harden deferred channel EOF/CLOSE replay after rekey

mjc:
## Summary
- replay deferred EOF/CLOSE in `flush_all_pending()` after pending data is flushed (the bug)
- add `#[must_use]` to `ChannelFlushResult` so the compiler catches any future caller that drops the result without handling pending controls
- refactor `flush_all_pending()` to delegate to `flush_pending()` per channel — single code path for data flush + control dispatch, making the bug class structurally impossible to reintroduce
- extract a small helper for taking deferred channel control flags
- add tests for `flush_pending` (single-channel, incomplete flush, idempotency) and multi-channel `flush_all_pending`

## Minimum fix

The commit **`fcc62bd` fix: replay deferred channel control after flush** is the smallest self-contained fix. Everything else in this PR is hardening and test coverage on top of it. The diff:

```diff
-    pub(crate) fn complete(wrote: usize, channel: &ChannelParams) -> Self {
+    pub(crate) fn complete(wrote: usize, channel: &mut ChannelParams) -> Self {
         ChannelFlushResult::Complete {
             wrote,
-            pending_eof: channel.pending_eof,
-            pending_close: channel.pending_close,
+            pending_eof: std::mem::take(&mut channel.pending_eof),
+            pending_close: std::mem::take(&mut channel.pending_close),
         }
     }

     pub fn flush_all_pending(&mut self) -> Result<(), crate::Error> {
-        for channel in self.channels.values_mut() {
-            Self::flush_channel(&mut self.write, channel)?;
+        let mut completed_channels = Vec::new();
+        for (&channel_id, channel) in self.channels.iter_mut() {
+            let flush_result = Self::flush_channel(&mut self.write, channel)?;
+            if matches!(flush_result, ChannelFlushResult::Complete { .. }) {
+                completed_channels.push((channel_id, flush_result));
+            }
+        }
+        for (channel_id, flush_result) in completed_channels {
+            self.handle_flushed_channel(channel_id, flush_result)?;
         }
         Ok(())
     }
```

The old `flush_all_pending` called `flush_channel` but discarded the result, so `handle_flushed_channel` was never called and `pending_eof`/`pending_close` were silently dropped. The old `complete()` also copied the flags without clearing them, which would have caused double-send if the result had ever been used — so that is fixed here too.

## OpenSSH parity

This fix matches the behaviour of OpenSSH's channel state machine. In
[`channels.c:3032-3045`](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/channels.c#L3032-L3045),
`channel_output_poll_input_open` calls
[`chan_ibuf_empty()`](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/nchan.c#L127-L146)
as soon as the input buffer drains — which in turn calls
[`chan_send_eof2()`](https://github.com/openssh/openssh-portable/blob/552a5c786b60a9cfe0d2c157dd18f78950529513/nchan.c#L182)
to send `SSH_MSG_CHANNEL_EOF` immediately after the last data byte. CLOSE
follows through the same state machine.

Before this fix, russh's `flush_all_pending()` (called on rekey completion)
sent queued channel data but silently dropped the deferred EOF/CLOSE flags,
violating that invariant.

### DIFF
--- a/russh/src/session.rs
+++ b/russh/src/session.rs
@@ -354,30 +354,19 @@ impl Encrypted {
     }
 
     pub fn flush_pending(&mut self, channel: ChannelId) -> Result<usize, crate::Error> {
-        let mut pending_size = 0;
-        let mut maybe_flush_result = Option::<ChannelFlushResult>::None;
-
-        if let Some(channel) = self.channels.get_mut(&channel) {
-            let flush_result = Self::flush_channel(&mut self.write, channel)?;
-            pending_size += flush_result.wrote();
-            maybe_flush_result = Some(flush_result);
-        }
-        if let Some(flush_result) = maybe_flush_result {
-            self.handle_flushed_channel(channel, flush_result)?
-        }
-        Ok(pending_size)
+        let flush_result = match self.channels.get_mut(&channel) {
+            Some(ch) => Self::flush_channel(&mut self.write, ch)?,
+            None => return Ok(0),
+        };
+        let wrote = flush_result.wrote();
+        self.handle_flushed_channel(channel, flush_result)?;
+        Ok(wrote)
     }
 
     pub fn flush_all_pending(&mut self) -> Result<(), crate::Error> {
-        let mut completed_channels = Vec::new();
-        for (&channel_id, channel) in self.channels.iter_mut() {
-            let flush_result = Self::flush_channel(&mut self.write, channel)?;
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #671, opened 2026-03-18
Update dev-dependencies (env_logger, clap, termion)

kpcyrd:
I noticed they are outdated, updating to the latest version.

(Also thanks for the 0.58.0 release!)


Eugeny:
Thanks!

PR #672, opened 2026-03-18
Reduce size of ReadSshIdBuffer, add unit tests

kpcyrd:
While reviewing the ssh-id handling code I noticed it's using a `Vec` for a 256 byte fixed-size heap buffer.

Since the Vec's `capacity` and `len` are never changed, we can use a `Box<[u8; 256]>` to reduce the size of ReadSshIdBuffer from `48 bytes` to `32 bytes`.

We could theoretically also put this on the stack, this would grow the ReadSshIdBuffer from `48 bytes` to `280 bytes` (not sure if we care), but it would further speed up the handshake.

Also adds a bunch of unit tests that I've used to ensure my change doesn't have any side-effects.

### DIFF
--- a/russh/src/ssh_read.rs
+++ b/russh/src/ssh_read.rs
@@ -6,10 +6,12 @@ use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, ReadBuf};
 
 use crate::Error;
 
+const SSH_ID_BUF_SIZE: usize = 256;
+
 /// The buffer to read the identification string (first line in the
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

kpcyrd:
Thanks for merging! :) Just to double-check, do you have any opinion on using stack or the heap for this? On glibc the stack is often 2-8MB, and on musl it's 128KB, so 256 bytes probably don't matter that much?

PR #684, opened 2026-03-31
fix(kex): separate GEX peer request validation from client config

DDtKey:
`russh` currently validates peer `diffie-hellman-group-exchange-*` requests the same way it validates client configuration. 

On the server side that is too strict: a client may legitimately advertise `min=1024`, and the server can still **safely choose a stronger group** as long as the client’s max still covers the server’s minimum expectation. [RFC 4419](https://datatracker.ietf.org/doc/html/rfc4419) also recommends `min=1024` and `max=8192`.

> Servers and clients SHOULD support groups with a modulus length of k
>   bits, where 1024 <= k <= 8192.  The recommended values for min and
>   max are 1024 and 8192, respectively.

If the goal is to disallow weak groups on the server side, the check that actually matters is `max >= 2048`, because that ensures the server can still choose a 2048+ DH group regardless of the peer’s lower min.

### Changes
- Split GexParams construction into explicit paths for client configuration and peer-request decoding.
  - Kept existing client-configuration behavior unchanged
- Relaxed peer-request validation so the server accepts requests whose advertised range still permits a 2048+ group.
- Added regression tests

### DIFF
--- a/russh/src/client/mod.rs
+++ b/russh/src/client/mod.rs
@@ -1685,22 +1685,42 @@ impl GexParams {
         preferred_group_size: usize,
         max_group_size: usize,
     ) -> Result<Self, Error> {
-        let this = Self {
+        Self::for_client_config(min_group_size, preferred_group_size, max_group_size)
+    }
+
+    pub fn for_client_config(
+        min_group_size: usize,
+        preferred_group_size: usize,
+        max_group_size: usize,
+    ) -> Result<Self, Error> {
+        Self::build(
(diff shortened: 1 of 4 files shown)


DDtKey:
A cleaner follow-up would be to separate the client and server validation paths more explicitly, but that felt out of scope for this fix.

Eugeny:
Looking good, thank you for the PR! @all-contributors please add @DDtKey for code

PR #694, opened 2026-04-23
Fix CI failures on main

mjc:
Fixes the failing CI run on `main`: https://github.com/Eugeny/russh/actions/runs/24744494719

- removes the unused legacy `HMAC_ORDER` constant left after excluding SHA-1 MACs from defaults
- pins coherent lower bounds for the RustCrypto prerelease stack so `cargo minimal-versions check --all-features --no-dev-deps` no longer selects mutually incompatible early RCs

Verified locally:
- `cargo fmt --check`
- `cargo clippy -- -D warnings`
- `cargo clippy --all-features -- -D warnings`
- `cargo minimal-versions check --all-features --no-dev-deps`

### DIFF
--- a/russh/Cargo.toml
+++ b/russh/Cargo.toml
@@ -28,7 +28,13 @@ serde = ["ssh-key/serde"]
 _bench = ["dep:criterion"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

PR #695, opened 2026-04-23
Reduce russh write-path copies with direct Bytes sends

mjc:
## Summary

This PR reduces write-path copying in `russh` by moving more packet assembly directly into `PacketWriter` and by sending owned channel payloads as `Bytes` where ordering allows.

For channel data, the new fast path writes `Bytes` payloads straight into `PacketWriter` instead of first staging plaintext into `Encrypted.write`. That path is used only when the channel is confirmed, no rekey is in progress, there is no older pending data for that channel, and the session write queue is empty. If any of those conditions are not met, behavior falls back to the existing queued path.

The same direct-writer path is now used when replaying pending channel data after a window adjustment or after rekey completion, again only when ordering permits. This keeps the existing ordering guarantees while avoiding the extra staging copy on the replay path too.

## API additions

This also adds owned-`Bytes` channel send helpers:

- `Channel::data_bytes`
- `Channel::extended_data_bytes`
- `ChannelWriteHalf::data_bytes`
- `ChannelWriteHalf::extended_data_bytes`

These helpers split payloads by writable window and max packet size using `Bytes::slice(...)`, so callers that already own a `Bytes` buffer can enqueue channel data without the extra copy inherent in the `AsyncWrite<&[u8]>` path.

## Packet writer changes

To support the direct path, `PacketWriter` is refactored so it can assemble and seal packets directly into its output buffer when compression is disabled, while still keeping a reusable plaintext packet buffer for retained-packet and compressed cases. The compressed path is also tightened so compressed output can be written directly into the final output buffer.

Retained KEXINIT-style packets are now kept as `Bytes`, and the stored client/server KEXINIT payloads in `Exchange` use `Bytes` as well.

## Hardening and validation

This PR also hardens channel write sizing by returning `Error::Inconsistent` when a peer advertises a zero max packet size instead of panicking.

Coverage was expanded around:

- direct channel-data writes
- pending-data replay through `PacketWriter`
- rekey replay behavior
- retained packet handling
- zlib + max-channel-packet-size behavior

## Wall-Clock Measurements

Temporary harnesses, not committed. Base is updated `main` at `c31cbc9`.

Numbers below are median-of-5 runs from the same local harness.

| case | main path | branch path | main median ns/op | branch median ns/op | faster by |
|---|---|---|---:|---:|---:|
| channel data, 32 KiB | current `Encrypted.write` staging | direct `Bytes` -> `PacketWriter` | 1,038.75 | 649.63 | 37.5% |
| channel data, 256 KiB | current `Encrypted.write` staging | direct `Bytes` -> `PacketWriter` | 10,418.58 | 5,690.90 | 45.4% |
| pending replay, 32 KiB | replay into `Encrypted.write` | replay into `PacketWriter` | 945.84 | 520.06 | 45.0% |
| pending replay, 256 KiB | replay into `Encrypted.write` | replay into `PacketWriter` | 9,649.69 | 4,508.97 | 53.3% |
| channel API, 32 KiB | `AsyncWrite<&[u8]>` copy path | owned `Bytes` API | 657.25 | 135.53 | 79.4% |
| channel API, 256 KiB | `AsyncWrite<&[u8]>` copy path | owned `Bytes` API | 6,256.34 | 953.55 | 84.8% |

## IAI/Callgrind Measurements

Same temporary harness. Instruction counts:

| case | main Ir | branch Ir | reduction |
|---|---:|---:|---:|
| channel data, 32 KiB | 82,811 | 48,104 | 41.9% |
| channel data, 256 KiB | 678,528 | 280,143 | 58.7% |
| pending replay, 32 KiB | 83,414 | 48,609 | 41.7% |
| pending replay, 256 KiB | 679,144 | 280,730 | 58.7% |

Retained packet construction is effectively flat by IAI: `67,817 Ir` on main vs `67,931 Ir` here for the 32 KiB retained packet case.

### DIFF
--- a/russh/src/channels/mod.rs
+++ b/russh/src/channels/mod.rs
@@ -359,7 +359,10 @@ impl<S: From<(ChannelId, ChannelMsg)> + Send + Sync + 'static> ChannelWriteHalf<
         Ok(())
     }
 
-    async fn reserve_writable_chunk(&self, remaining: usize) -> usize {
+    async fn reserve_writable_chunk(&self, remaining: usize) -> Result<usize, Error> {
+        if self.max_packet_size == 0 {
+            return Err(Error::Inconsistent);
+        }
         loop {
             let mut window_size = self.window_size.value.lock().await;
             let writable = (self.max_packet_size as usize)
@@ -370,10 +373,11 @@ impl<S: From<(ChannelId, ChannelMsg)> + Send + Sync + 'static> ChannelWriteHalf<
                 if *window_size > 0 {
                     self.window_size.notifier.notify_one();
                 }
-                return writable;
+                return Ok(writable);
             }
+            let notified = self.window_size.notifier.notified();
             drop(window_size);
-            self.window_size.notifier.notified().await;
+            notified.await;
         }
     }
 
@@ -384,7 +388,7 @@ impl<S: From<(ChannelId, ChannelMsg)> + Send + Sync + 'static> ChannelWriteHalf<
 
         let mut offset = 0;
         while offset < data.len() {
-            let writable = self.reserve_writable_chunk(data.len() - offset).await;
+            let writable = self.reserve_writable_chunk(data.len() - offset).await?;
             let end = offset + writable;
             let chunk = data.slice(offset..end);
             let msg = match ext {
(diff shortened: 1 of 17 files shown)


mjc:
if desired I can add the benchmark harness, also. just wasn't sure it would be wanted

Eugeny:
Thank you so much! It will take me a while to properly review since I'm currently away until the 1st - but from a cursory glance everything looks good :+1:

mjc:
Yeah, sorry about the size of it. if it helps I could split it into a bunch of smaller PRs for easier examination when you're back.

There are a couple ways to slice this particular pie that might make it simpler.

Eugeny:
Thanks a ton, great work :+1:

PR #710, opened 2026-05-21
fix: accept empty name-list in KEXINIT (RFC 4251 §5)

lovesegfault:
`NameList::from_encoded_string("")` previously failed with
`Error::CharacterEncoding` because `"".split(',')` yields one empty
element, which the per-name validation rejects.

RFC 4251 §5 permits a zero-length name-list. Servers that only offer
AEAD ciphers (observed: hssh_0.1.0.0) send empty `mac_algorithms`
fields, and `client::connect` would fail before the AEAD-needs-no-MAC
fallback in `negotiation.rs` could fire.


Eugeny:
Thank you!

PR #714, opened 2026-05-24
Send USERAUTH_BANNER after first USERAUTH_REQUEST instead of with SERVICE_ACCEPT

Dean6767:
## Problem

Currently, `server_accept_service` emits `SSH_MSG_USERAUTH_BANNER` immediately after `SERVICE_ACCEPT`, before the client has sent its first `USERAUTH_REQUEST`.

Many SSH clients (notably **OpenSSH**) only display banners that arrive *after* the first `USERAUTH_REQUEST`. As a result, when using `Config::auth_banner`, the banner is sent on the wire but **silently ignored by OpenSSH clients** — the user never sees it.

You can verify this with:
```bash
ssh -v user@host 2>&1 | grep -i banner
```
With current `russh`, no `Authentication request banner length: N` line ever appears, regardless of `auth_banner` being set.

## Solution

Move the `USERAUTH_BANNER` emission into `server_read_auth_request`, gated on `rejection_count == 0` so it only fires on the very first auth request. This matches the behavior of [`golang.org/x/crypto/ssh`](https://pkg.go.dev/golang.org/x/crypto/ssh#ServerConfig.BannerCallback), which has been widely deployed without compatibility issues.

## Testing

Tested against:
- OpenSSH 7.6 / 8.x client — banner now displayed correctly before password prompt
- Trilead SSH2 Java library — works as before
- The fix is minimal (~13 lines changed) and preserves the existing `Config::auth_banner: Option<&'static str>` API

## Diff size

1 file changed, 23 insertions(+), 10 deletions(-)


Eugeny:
Thanks! You need to retarget the PR to the `main` branch. Also, tests are failing, presumably due to the changes in the message timing

PR #719, opened 2026-05-30
fix: support SEC1 EC keys with full domain parameters

### LINKED ISSUE #718: Bug: All PEM/SEC1-format EC private keys fail to parse — TagUnexpected { expected: SEQUENCE, actual: OCTET STRING }
## Description

Any EC private key in traditional PEM format (`-----BEGIN EC PRIVATE KEY-----`) fails to parse with a DER decoding error. This affects **all** EC key types (nistp256, nistp384, nistp521) regardless of key size. Keys in OpenSSH format (`-----BEGIN OPENSSH PRIVATE KEY-----`) work correctly.

RSA and Ed25519 keys are unaffected in any format.

## Steps to Reproduce

```bash
# 1. Generate any EC key in PEM format
ssh-keygen -t ecdsa -b 256 -f /tmp/test_pem -m PEM -N ""

# 2. Parse with russh
```

```rust
use russh_keys::decode_secret_key;
let pem = std::fs::read_to_string("/tmp/test_pem").unwrap();
let _key = decode_secret_key(&pem, None).unwrap();
// Error: Keys(Der(Error { kind: TagUnexpected { expected: Some(Tag(0x10: SEQUENCE)), actual: Tag(0x04: OCTET STRING) }, position: None }))
```

Contrast with OpenSSH format (works fine):

```bash
ssh-keygen -t ecdsa -b 256 -f /tmp/test_openssh -N ""  # default format
# → Parses successfully
```

## Expected Behavior

`decode_secret_key` should successfully parse any valid SEC1-encoded EC private key. The DER content is structurally correct — the first byte is `0x30` (SEQUENCE).

## Actual Behavior

The error originates from `decode_pkcs8` (`russh/src/keys/format/pkcs8.rs:27`):

```rust
match doc.decode_msg::<sec1::EcPrivateKey>() {
    Ok(key) => { /* never reached for PEM-format EC keys */ }
    Err(_) => { /* falls through to PKCS#8, which also fails */ }
}
```

## Verification

I verified the DER bytes are correct by decoding the base64 independently:

```
PEM header:  -----BEGIN EC PRIVATE KEY-----
DER bytes:   30 82 02 9d 02 01 01 04 42 ...
             ↑ SEQUENCE (correct at offset 0)
```

The DER structure is a valid SEC1 `ECPrivateKey`:

```
SEQUENCE {
    INTEGER 1                    (version)
    OCTET STRING (66 bytes)      (private key, for P-521)
    [1] BIT STRING               (public key)
}
```

Yet the `sec1::EcPrivateKey` decoder reports seeing `0x04` (OCTET STRING) at position 0, when the actual byte is `0x30` (SEQUENCE). The decoder appears to be reading from an incorrect offset.

## Affected

- `sec1` crate `0.8` (with `der` feature)
- `der` crate `0.8`
- Tested with russh `0.60.1` (npm `russh@0.1.37`)

**Not affected:**
- RSA keys (all formats)
- Ed25519 keys
- EC keys in OpenSSH format (`-----BEGIN OPENSSH PRIVATE KEY-----`)

## Workaround

Convert PEM EC keys to OpenSSH format:

```bash
ssh-keygen -p -f ~/.ssh/id_ecdsa -m openssh
```

## Root Cause

In `decode_pkcs8`, the call to `doc.decode_msg::<sec1::EcPrivateKey>()` fails on valid SEC1 DER. The `sec1` crate ASN.1 decoder appears to read from the wrong offset — it sees the inner OCTET STRING (at byte 7) instead of the outer SEQUENCE (at byte 0). This affects all PEM-format EC keys regardless of curve or size.

SakuraPuare:
## Problem

SEC1 EC private keys that use full domain parameters (instead of a named curve OID) fail to parse with:

```
Keys(Der(Error { kind: TagUnexpected { expected: Some(Tag(0x10: SEQUENCE)), actual: Tag(0x04: OCTET STRING) }, position: None }))
```

This affects keys generated with `openssl ecparam -param_enc explicit` and similar tools that embed the full EC domain parameters in the `[0]` parameters field of the SEC1 structure.

Fixes #718

## Root Cause

The `sec1` crate's `EcPrivateKey` decoder expects the `[0]` parameters field to contain either a named curve OID or nothing (optional field). But some key generators produce a full EC domain parameters SEQUENCE instead, which the `sec1` crate rejects with a `TagUnexpected` error.

## Fix

Added a manual DER parser as a fallback in `decode_pkcs8` that:

1. Detects SEC1 keys with full domain parameters (when `sec1::EcPrivateKey` decode fails)
2. Extracts the curve OID from the prime field length in the domain parameters
3. Constructs the EC keypair directly from the private key bytes using `p256`/`p384`/`p521` crates

The fallback only runs when the `sec1` crate's strict parser fails, so there is no performance impact on normal keys.

## Changes

- `russh/src/keys/format/pkcs8.rs`: Added fallback parser functions for SEC1 keys with full domain parameters
- `russh/src/keys/format/tests.rs`: Added tests for P-256 and P-521 keys with explicit domain parameters, and negative tests for malformed DER

## Testing

- 165 tests pass, 0 warnings
- New tests cover P-256 and P-521 with explicit domain parameters
- Negative tests for malformed DER and truncated keys
- All existing tests continue to pass

### DIFF
--- a/russh/src/keys/format/pkcs8.rs
+++ b/russh/src/keys/format/pkcs8.rs
@@ -34,7 +34,12 @@ pub fn decode_pkcs8(
             Ok(PrivateKey::new(KeypairData::Ecdsa(kp), "")?)
         }
         Err(_) => {
-            // ASN.1 key
+            // SEC1 key with full domain parameters (not a named curve OID)
+            match decode_sec1_with_full_domain_params(ciphertext) {
+                Ok(kp) => return Ok(PrivateKey::new(KeypairData::Ecdsa(kp), "")?),
+                Err(_) => {},
+            }
+            // ASN.1 key (PKCS#8)
             Ok(
                 pkcs8_pki_into_keypair_data(doc.decode_msg::<PrivateKeyInfoRef<'_>>()?)?
                     .try_into()?,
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you! I've updated the code to use the `der` crate for parsing

Eugeny:
@all-contributors please add @SakuraPuare for code

PR #726, opened 2026-06-17
pageant: fall back to GetUserNameA when GetUserNameExA fails

cwatanab:
## Problem

`GetUserNameExA(NameUserPrincipal)` requires a UPN (User Principal Name), which does not exist on non-domain-joined Windows machines (standalone/workgroup). On such machines the API fails with `ERROR_NONE_MAPPED` (0x80070534), preventing any named-pipe connection to a running Pageant instance.

This affects applications like DBX (Database Explorer) and the `pageant` crate itself.

## Root Cause

The original PuTTY Pageant (C implementation) falls back to `GetUserNameA` when `GetUserNameExA` fails. The code comment acknowledges this but the Rust port chose not to implement the fallback, under the assumption that all modern Windows versions support `GetUserNameExA`. However, even on modern Windows 10/11, a non-domain-joined machine has no UPN configured, and the API call fails.

## Fix

Fall back to the `USERNAME` environment variable when `GetUserNameExA(NameUserPrincipal)` fails. This provides the SAM account name, which matches what the original PuTTY Pageant's `GetUserNameA` fallback would return on local accounts.

## Testing

- Verified on Windows 10 (non-domain-joined workstation) where `GetUserNameExA(NameUserPrincipal)` consistently fails with `ERROR_NONE_MAPPED`
- Verified that `USERNAME` env var contains the correct SAM account name
- The fallback produces the same username as what the user would expect

This is consistent with how the reference PuTTY Pageant implementation handles this edge case.

### DIFF
--- a/pageant/Cargo.toml
+++ b/pageant/Cargo.toml
@@ -45,4 +45,5 @@ namedpipes = [
     "dep:windows-strings",
     "windows/Win32_Security_Authentication_Identity",
     "windows/Win32_Security_Cryptography",
+    "windows/Win32_System_WindowsProgramming",
 ]
--- a/pageant/src/namedpipes.rs
+++ b/pageant/src/namedpipes.rs
@@ -65,10 +65,13 @@ impl PageantStream {
                 Some(PSTR(name_buf.as_mut_ptr())),
                 &mut name_length,
             ) {
-                // Pageant falls back to GetUserNameA here,
-                // but as far as I can tell, all Versions of Windows supported by Rust today
-                // should be able to answer the UserNameEx request - the comments in Pageant source
-                // point to Windows XP and earlier compatibility...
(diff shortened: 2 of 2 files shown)


rebelliongeeks:
Can confirm on Windows 11 with a local (non-domain) account: GetUserNameExA(NameUserPrincipal) fails with 0x80070534, so the transport never gets a pipe name even though the pipe is right there as `pageant.<USERNAME>.<hash>`.

The USERNAME fallback would've produced the correct name for us.

Hit this while evaluating russh 0.62.6 for a desktop app. +1 for merging.

Eugeny:
Thanks - I've updated it to fall back to GetUserNameA to match what putty is actually doing

Eugeny:
@all-contributors add @cwatanab for code

PR #729, opened 2026-06-23
fix(server): drain Handle::data backlog before select!

Yaminyam:
## Summary
`tokio::select!` in the server session loop doesn't always wake up promptly for messages produced by spawned tasks via `Handle::data()` when the internal mpsc channel is being filled faster than the select cycle drains it. In interactive PTY servers where a separate task pumps shell output through `Handle::data()`, this shows up as the client missing chunks of output until some other event happens to wake select.

This PR drains the receiver with `try_recv()` before entering `select!`, with a 64-message cap so select still gets to run between batches and client input (Ctrl+C, window resize, …) stays responsive.

## Reproduction
Any russh-based SSH server with a PTY session whose shell output is generated from a spawned task pushing via `Handle::data()` faster than the session loop's select cycle reads from its mpsc channel — e.g. `yes | head -n 100000` over an interactive PTY. The client sees periodic stalls / dropped output until further activity.

## Change
In `run_inner` (server session loop), before the existing `tokio::select!`, batch-drain up to `MAX_MESSAGES_PER_BATCH = 64` messages from `self.receiver` with `try_recv()` and dispatch through the same handlers `select!` would have used: `data` / `extended_data` / `eof` / `close` / `channel_success` / `channel_failure` / `xon_xoff_request` / `exit_status_request` / `exit_signal_request` / `WindowAdjusted` / `ChannelOpen{Agent,Session,DirectTcpip,...}`. Gated on `!self.kex.active()` so the rekey state machinery isn't perturbed.

The cap exists so high-throughput output can't starve client input — once a batch is full we fall through to `select!`, which picks up any client-side event (input, channel close, keepalive timeout) before the next batch.

## Origin
This is the same fix [bssh](https://github.com/lablup/bssh) has been carrying as `crates/bssh-russh/patches/handle-data-fix.patch`. Its README pins at [`inureyes/russh:fix/handle-data-from-spawned-tasks`](https://github.com/inureyes/russh/tree/fix/handle-data-from-spawned-tasks) and notes "When merged upstream, this fork will be deprecated" — this PR is the upstream half of that contract. Original author of the fix: @inureyes (bssh maintainer).

## Test plan
- [x] `cargo build -p russh` → clean
- [ ] Manual PTY soak test recommended before merge: spawn-task-driven `Handle::data()` server, confirm no output stalls under `yes | head -n 100000`-style load.

### DIFF
--- a/russh/src/server/session.rs
+++ b/russh/src/server/session.rs
@@ -6,7 +6,7 @@ use channels::WindowSizeRef;
 use kex::ServerKex;
 use log::debug;
 use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
-use tokio::sync::mpsc::{Receiver, Sender, channel};
+use tokio::sync::mpsc::{Receiver, Sender, channel, error::TryRecvError};
 use tokio::sync::oneshot;
 
 use super::*;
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks for bringing it to my attention - I've added my own cleaner fix in #731

@all-contributors please add @inureyes for code

PR #730, opened 2026-06-26
fix: gate WINDOW_ADJUST on consumption to fix per-channel HOL blocking

lovesegfault:
ran into this and after some debugging with claude this was the result, submitting mostly for visibility since the patch ended up bigger than i expected.

The session loop forwarded incoming `CHANNEL_DATA` to the per-channel
mpsc with `chan.send().await` and replenished the SSH receive window on
receipt. A consumer that stopped draining one channel parked the entire
connection — every other channel, keepalives, and the kex path included.

This replaces the blocking send with a `ChannelBacklog`: `try_send` to
the mpsc, queue on `Full`, and withhold `WINDOW_ADJUST` while a backlog
exists so the peer's window for that channel drains to zero and *it*
stops sending — while the loop stays live for everything else. The queue
is bounded by the SSH window for compliant peers (RFC 4254 §5.2); a peer
that ignores flow control is the same authenticated-DoS exposure as
before.

Summary of changes:

- `Encrypted::adjust_window_size` is split into `record_received_data`
  (debit on receipt) and `replenish_receive_window` (emit
  `WINDOW_ADJUST` once the channel has no backlog)
- a `drain_notify: Arc<Notify>` lets each `ChannelReadHalf` recv wake
  the loop to flush its backlog and re-open the window; the drain arm is
  gated on `!kex.active()` so it never emits mid-rekey
- `CHANNEL_CLOSE`/`OPEN_FAILURE` defer removing the `ChannelRef` until
  the backlog has drained, so a slow consumer still receives the
  buffered tail (Data/Eof/ExitStatus/Close) before seeing `None`
- `WindowAdjusted` is never queued (informational; the `WindowSizeRef`
  side effect already happened); all remaining `.send().await` sites
  (the eight server `CHANNEL_REQUEST` arms, `Open`) are routed through
  the non-blocking forward as well
- `ChannelRx` latches EOF locally instead of `receiver.close()` so
  post-EOF backlog can still drain into the mpsc
- `Session::drop` does a final best-effort drain so the buffered tail
  reaches consumers on every loop-exit path

Side effect: client `replenish_receive_window` now uses
`target_window_size` like the server already did, so a client
`Handler::adjust_window` override actually takes effect on the wire
(previously it was stored but ignored).

Handlers that consume via `Handler::data()` rather than
`Channel::wait()` should drop the `Channel` (or its read half) so the
window keeps re-opening; `examples/test.rs` is updated and the
constraint is documented on the trait method.

`tests/test_per_channel_flow_control.rs` covers the regression: it opens
two channels, parks one's consumer, floods it past mpsc capacity until
its window hits zero, and asserts the sibling is still served and the
stalled channel resumes (via the drain-path `WINDOW_ADJUST`) once its
consumer catches up.

### DIFF
--- a/russh/examples/test.rs
+++ b/russh/examples/test.rs
@@ -31,7 +31,7 @@ async fn main() -> anyhow::Result<()> {
 
 #[derive(Clone)]
 struct Server {
-    clients: Arc<Mutex<HashMap<(usize, ChannelId), Channel<Msg>>>>,
+    clients: Arc<Mutex<HashMap<(usize, ChannelId), ChannelWriteHalf<Msg>>>>,
     id: usize,
 }
 
@@ -56,8 +56,12 @@ impl server::Handler for Server {
     ) -> Result<(), Self::Error> {
         {
             debug!("channel open session");
+            // Only the write half is stored: data is consumed via the
+            // `Handler::data` callback below, so the read half is dropped to
+            // keep the SSH receive window open.
+            let (_read, write) = channel.split();
             let mut clients = self.clients.lock().unwrap();
-            clients.insert((self.id, channel.id()), channel);
+            clients.insert((self.id, write.id()), write);
         }
         reply.accept().await;
         Ok(())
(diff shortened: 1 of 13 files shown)


Eugeny:
This is just due to incorrect library use - once you get an incoming `Channel`, you're responsible for draining its event loop with `Channel::wait()`. If you don't it will cause the internal buffer fillup you're describing

PR #734, opened 2026-07-06
Fix deadlock when handling open channel on contended handler

EpicEric:
Before #723, synchronous channel acceptance/rejection would happen regardless of message dispatch. Now, the channel reply uses the same bounded mpsc channel that the handler consumes from.

However, under contention (for example, with multiple simultaneous channels opening and writing over the same handler), the awaits in `ChannelOpenHandleInner::accept` and `ChannelOpenHandleInner::reject` may hang since the consumer is the handler itself, thus leading to a self-deadlock.

This PR changes channel confirmation to use its own unbounded channel on both server and client implementations, avoiding any awaits (since send is sync), and thus, any deadlocks under load.

Because of the change to `ChannelOpenHandleInner::accept` and `ChannelOpenHandleInner::reject` into sync methods, this is backwards-incompatible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ async-trait = "0.1.50"
 base16ct = "1"
 byteorder = "1.4"
 bytes = "1.7"
-digest = "0.11.0-rc.5"
+digest = "0.11.0"
 delegate = "0.13"
 env_logger = "0.11"
 futures = "0.3"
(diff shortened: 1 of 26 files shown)


EpicEric:
I've added a test_contention regression test that sometimes locks up on russh 0.62.1 due to the race condition, but runs correctly with these changes.

EpicEric:
Despite the accept/reject methods being fully sync, I could change them back to async as to not break backwards compatibility.

Eugeny:
Thanks for the PR! Please change the accept/reject fns back to async - it's barely an issue for client code, and I'd like to avoid a breaking change for a bugfix

PR #740, opened 2026-07-20
Make RunningServer be pub constructable

zaczkows:
- since it's possible to overwrite `Server::run_on_socket`, which returns `RunningServer<>`, the latter must be publicly constructable


Eugeny:
`run_on_socket` is provided, but not intended for reimplementation. What would the usecase be for overriding if you're already writing your own server run method anyway?

PR #742, opened 2026-08-04
cryptovec: add Motor OS support

lasiotus:
russhd in Motor OS uses russh: https://github.com/moturus/motor-os/tree/main/src/bin/russhd


Eugeny:
Thanks!

PR #750, opened 2026-08-17
fix: don't truncate decompressed packets that expand more than 2x

### LINKED ISSUE #749: zlib decompression truncates packets
When compression is on, russh loses the tail of any packet whose data compresses better than 2:1. The packet comes out short, fails to decode, and russh closes the connection. An OpenSSH client with `Compression yes` dies on the first packet that compresses well.

russh's own compressor and decompressor can't round-trip one packet:

```rust
use russh::compression::{Compress, Compression, Decompress};

fn main() {
    let mut comp = Compress::None;
    let mut decomp = Decompress::None;
    Compression::Zlib.init_compress(&mut comp);
    Compression::Zlib.init_decompress(&mut decomp);

    let payload: Vec<u8> = (0..37).map(|i| b"abcdefgh"[i % 8]).collect();

    let mut cbuf = Vec::new();
    let compressed = comp.compress(&payload, &mut cbuf).unwrap().to_vec();

    let mut dbuf = Vec::new();
    let out = decomp.decompress(&compressed, &mut dbuf).unwrap();

    assert_eq!(out.len(), payload.len(), "sent 37 bytes, got {} back", out.len());
}
```

This panics with `sent 37 bytes, got 36 back` on 0.62.6. Sweeping sizes 1..=4096, 4049 come back short. I tracked this down together with Claude, which also put together a standalone repro repo: https://github.com/luizribeiro/russh-decompress-repro

**Affected:** 0.61.1 through 0.62.6, and `main`. Introduced by 2692a6f6, the fix for [GHSA-wwx6-x28x-8259](https://github.com/Eugeny/russh/security/advisories/GHSA-wwx6-x28x-8259).

## Cause

In `Decompress::decompress` (`russh/src/compression.rs`), the loop that grows the output buffer decides when to stop using counters read *before* the current `z.decompress` call:

```rust
let n_in_ = z.total_in() as usize - n_in;    // read BEFORE the call
let n_out_ = z.total_out() as usize - n_out;
let d = z.decompress(&input[n_in_..], &mut output[n_out_..], flush);
match d? {
    flate2::Status::Ok | flate2::Status::BufError => {
        let consumed_all_input = n_in_ == input.len();  // stale
        let output_full = n_out_ == output.len();       // stale
        if !output_full && consumed_all_input {
            break;
        }
```

In the example above (18 bytes compressed, buffer starts at 18): call 1 fills the buffer and consumes all input, call 2 makes no progress so the buffer grows to 36, call 3 writes 18 more bytes and fills it exactly — but the stale check compares 18 to 36, decides the buffer is not full, and breaks with one byte still inside zlib. The packet is returned short, and the leftover byte corrupts the stream for every later packet.

The tests added with 2692a6f6 miss this because `ZlibEncoder::finish()` produces a finished stream, which leaves the loop through `StreamEnd`; real SSH packets are partial flushes and leave through the stale checks.

## Fix

Read the counters after the call:

```diff
-let consumed_all_input = n_in_ == input.len();
-let output_full = n_out_ == output.len();
+let consumed_all_input = z.total_in() as usize - n_in == input.len();
+let output_full = z.total_out() as usize - n_out == output.len();
```

With this change all 4096 sizes round-trip, and OpenSSH + git over a russh server works again with `Compression yes`. Happy to send a PR with the fix and a regression test.

luizribeiro:
Fixes #749.

`Decompress::decompress` checked its loop-exit conditions using progress counters read before the current inflate call. A drain call that filled the output buffer exactly would break out of the loop with data still pending inside zlib: the packet came back short, and the leftover bytes corrupted the stream for every later packet. Any packet compressing better than 2:1 hit this, which killed every compressed session (e.g. OpenSSH clients with `Compression yes`).

This reads the counters after the call instead, and adds a regression test that round-trips partial-flush packets on a continuous stream — the shape real SSH traffic has, which the existing finished-stream tests don't cover. The test fails without the fix (`sent 37 bytes, got 36 back`) and passes with it; the full suite passes.

### DIFF
--- a/russh/src/compression.rs
+++ b/russh/src/compression.rs
@@ -216,6 +216,26 @@ mod tests {
         let decompressed = decompressor.decompress(&compressed, &mut output).unwrap();
         assert!(decompressed.is_empty());
     }
+
+    #[test]
+    fn partial_flush_packets_round_trip() {
+        // Real SSH packets are partial flushes on one continuous stream, not
+        // a finished stream, and routinely decompress to more than twice
+        // their compressed size.
+        let mut comp = Compress::None;
+        let mut decomp = Decompress::None;
+        Compression::Zlib.init_compress(&mut comp);
+        Compression::Zlib.init_decompress(&mut decomp);
+
+        for len in 1..=4096usize {
+            let payload: Vec<u8> = (0..len).map(|i| b"abcdefgh"[i % 8]).collect();
+            let mut cbuf = Vec::new();
+            let compressed = comp.compress(&payload, &mut cbuf).unwrap().to_vec();
+            let mut dbuf = Vec::new();
+            let out = decomp.decompress(&compressed, &mut dbuf).unwrap();
+            assert_eq!(out, payload.as_slice(), "payload of {len} bytes came back wrong");
+        }
+    }
 }
 
 #[cfg(feature = "flate2")]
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! @all-contributors add @luizribeiro for code

PR #752, opened 2026-08-17
Support host certificates in client

Eugeny:
#748 + fixes


oni303:
Would this also add support for servers to authenticate users that provide a signed certificate.
Please have a look at #641 we worked and tested quite a bit

Eugeny:
@oni303 no worries - I've seen all 3 PRs - I'll merge client support here since and then merge your server support on top with a few changes

PR #757, opened 2026-08-22
fix: don't truncate compressed packets that outgrow the output reserve

luizribeiro:
The existing code assumes a compressed packet is always smaller than the uncompressed one: `Compress::compress_into` sizes zlib's output buffer at `input.len() + 10` bytes and works from there.

That isn't true when the data is already compressed. For pack files, encrypted blobs or media, zlib has nothing left to squeeze out, so it wraps the bytes in a stored block and hands back slightly *more* than it was given. Past 32 KiB the stored-block overhead is more than the 10 bytes of slack, so the buffer is too small.

Overflowing is supposed to be recoverable: the loop grows the buffer and calls zlib again. But it only grows on `flate2::Status::BufError`, and zlib returns `BufError` only when it is completely stuck. Filling the buffer while still making progress comes back as `Status::Ok`, which the loop reads as "done", so it sends the packet with the tail still inside zlib. The packet arrives shorter than its header claims, and the bytes left behind are prepended to the next one, so the stream stays broken from there on. OpenSSH reports it as `channel 0: get data: incomplete message`: a `git clone` against a russh server dies as soon as pack data starts flowing, while ordinary traffic is unaffected because it does compress.

The fix stops asking zlib whether it is stuck and asks whether it left any room. Room to spare means it emitted everything; a buffer filled to the brim means there may be more, so grow and call again. `compress` now delegates to `compress_into` so the two copies of the loop can't drift apart again.

The new tests round-trip packets that outgrow the reserve, including through `compress_into` at a non-zero `start_len`, the shape the packet writer uses. The existing tests only cover compressible payloads under 4096 bytes, which always fit. The new ones fail without the fix (`payload of 32769 bytes came back as 32735 bytes`) and pass with it; the full suite passes. Verified end to end too: a stock `git clone` over a compressed session against a russh-backed git server fails on 0.62.6 with exactly that error and succeeds with this change.

### DIFF
--- a/russh/src/compression.rs
+++ b/russh/src/compression.rs
@@ -236,6 +236,77 @@ mod tests {
             assert_eq!(out, payload.as_slice(), "payload of {len} bytes came back wrong");
         }
     }
+
+    /// Deterministic bytes with no redundancy for deflate to exploit, so the
+    /// packet comes back out larger than it went in.
+    fn incompressible(seed: u64, len: usize) -> Vec<u8> {
+        let mut state = seed
+            .wrapping_mul(6364136223846793005)
+            .wrapping_add(1442695040888963407);
+        (0..len)
+            .map(|_| {
+                state ^= state << 13;
+                state ^= state >> 7;
+                state ^= state << 17;
+                (state >> 24) as u8
+            })
+            .collect()
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! I've accidentally started working on it myself having missed your PR - but I've credited you for the fix anyway

PR #764, opened 2026-09-03
fix(client): encode the negotiated hash algorithm for RSA certificates

inureyes:
`Method::FutureCertificate` carries a `hash_alg`, but both the probe request and the signed request encoded `cert.algorithm().to_certificate_type()`, which always yields `ssh-rsa-cert-v01@openssh.com` for an RSA certificate. A server that rejects SHA-1 signatures therefore refuses RSA user certificates even when the client selected SHA-256 or SHA-512.

`client_make_to_sign` now takes the certificate hash algorithm and applies it with `with_hash_alg` in both places, so the wire carries `rsa-sha2-512-cert-v01@openssh.com` or `rsa-sha2-256-cert-v01@openssh.com` when one is selected. Non-RSA certificates encode identically to before, and the self-signed `OpenSshCertificate` path passes `None`, so its behavior is unchanged.

A new test drives all three hash selections through both the probe and the signed request. Removing either `with_hash_alg` call makes it fail with `ssh-rsa-cert-v01@openssh.com`.

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -255,6 +255,7 @@ impl Session {
                                             key: key.clone(),
                                             hash_alg,
                                         },
+                                        None,
                                         &mut self.common.buffer,
                                     )?;
                                     let len = self.common.buffer.len();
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! @all-contributors add @inureyes for code

PR #766, opened 2026-09-04
Fix enforce rekey read and time limits

inureyes:
## Description

`Limits` exposes `rekey_read_limit` and `rekey_time_limit`, but the only place that consults them is `Encrypted::flush`, which is reached on the write path. A session that mostly downloads, or one that idles, keeps the same keys no matter how the limits are configured. `rekey_write_limit` is also off by the current epoch's cleartext KEX packets, because `SSHBuffer::bytes` is never reset when a cipher is installed.

- `cipher::read` counts inbound payload bytes, and the client and server loops compare that count against `rekey_read_limit`, resetting it when a key exchange completes.
- `PacketWriter::set_cipher` resets the write counter so each key epoch starts at zero.
- Both loops arm a deadline from `rekey_time_limit` and `Encrypted::last_rekey`, so an idle session still rekeys on time.
- Automatic rekeying is gated on completed authentication. Without that, a small `RekeyLimit` makes pre-authentication traffic start a second key exchange while the first is still being set up. On the server, `USERAUTH_SUCCESS` is flushed while the state is still `InitCompression`, so deferred server compression is activated before `KEXINIT` when rekeying is the first post-authentication write.

`Session::initiate_rekey` and `Msg::Rekey` are unchanged and still rekey on demand.

Ten tests cover the read limit, the deadline, the pre- and post-authentication gate on both sides, and the epoch reset. Dropping the byte accounting makes the two `sshbuffer` tests fail.

...

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/russh/src/cipher/mod.rs
+++ b/russh/src/cipher/mod.rs
@@ -345,10 +345,12 @@ pub(crate) async fn read<R: AsyncRead + Unpin>(
         .len()
         .checked_sub(padding_length)
         .ok_or(Error::IndexOutOfBounds)?;
+    let payload_len = plaintext_end.saturating_sub(PADDING_LENGTH_LEN);
 
     // Sequence numbers are on 32 bits and wrap.
     // https://tools.ietf.org/html/rfc4253#section-6.4
     buffer.seqn += Wrapping(1);
+    buffer.bytes = buffer.bytes.saturating_add(payload_len);
     buffer.len = 0;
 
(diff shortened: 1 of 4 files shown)


inureyes:
`Build-WASM` is fixed in 4067314. The `Clippy` and `Minimal-versions` failures predate this PR and come from `main` on the 1.91 toolchain. #768 fixes both, if that helps.

PR #767, opened 2026-09-04
Add support hostkeys-prove-00@openssh.com

inureyes:
## Description

OpenSSH servers advertise their full host key set with `hostkeys-00@openssh.com` and expect the client to challenge it with `hostkeys-prove-00@openssh.com` before adopting any of the keys. Without that exchange a client cannot implement `UpdateHostKeys` safely, because an announced key is only trustworthy once the server proves it holds the matching private key. russh has no way to send the request today.

`Session::request_hostkeys_prove` sends the global request with `want_reply` set and one blob per candidate key. `REQUEST_SUCCESS` decodes every signature in order into a `HostKeysProof` that also carries the session ID. `REQUEST_FAILURE`, a malformed body, and an unencrypted session all resolve to `None`.

The proof is returned raw. Verifying each signature over `"hostkeys-prove-00@openssh.com" || session_id || key_blob` stays with the caller, which is where the policy for adopting a new key lives anyway.

#759 proposes a generic custom-global-request API. If that lands, this could be rewritten on top of it, though the response decoding would still need somewhere to live.

...

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/russh/src/client/encrypted.rs
+++ b/russh/src/client/encrypted.rs
@@ -962,6 +962,26 @@ impl Session {
                         map_err!(ensure_end(&r))?;
                         let _ = return_channel.send(true);
                     }
+                    Some(GlobalRequestResponse::HostKeysProve {
+                        return_channel,
+                        session_id,
+                    }) => {
+                        let mut signatures = Vec::new();
+                        while !r.is_empty() {
+                            match Bytes::decode(&mut r) {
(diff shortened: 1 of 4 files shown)


inureyes:
The `Clippy` and `Minimal-versions` failures predate this PR and come from `main` on the 1.91 toolchain. I submitted #768 to fix both.

PR #768, opened 2026-09-04
Fix Clippy and minimal-versions CI failures on Rust 1.91

inureyes:
## Description

`Clippy` and `Minimal-versions` have been failing on `main` since the 1.91 toolchain reached the runners, independently of any pull request: https://github.com/Eugeny/russh/actions/runs/33791529793 at 8a51a0d.

Clippy 1.91 extended `collapsible_if` to nested `if let`, which flags 40 sites. `cargo clippy --fix` rewrites them as let-chains, but neither stable rustfmt nor nightly 1.10.0 reformats the result, so every body keeps its pre-collapse indentation until it is fixed by hand. That churn lands in the session and encrypted modules that most open PRs touch, so this allows the lint next to the two allows already in `lib.rs` rather than reformatting them now.

Under minimal versions, `yasna`'s `num-bigint = "0.4"` resolves to 0.4.0, whose `bits().div_ceil(&u64::from(bits))` now picks the inherent `u64::div_ceil` over `Integer::div_ceil` and fails to compile. 0.4.6 switched those three call sites to UFCS, so the floor is raised with a renamed optional dependency tied to the same feature as `yasna`, the same way `cipher`, `ghash`, `polyval` and `universal-hash` are already pinned in this file.

Verified with the CI commands: `cargo clippy` with and without `--all-features`, `cargo minimal-versions check --all-features --no-dev-deps`, `cargo fmt --check`, `cargo test --all-features`, plus the 1.89.0 `wasm32-wasip1` build and an MSRV check.

## AI Usage

* [ ] Fully vibe coded
* [x] AI-designed, AI-coded, manually checked
* [ ] Human-designed, AI-coded

### DIFF
--- a/russh/Cargo.toml
+++ b/russh/Cargo.toml
@@ -17,7 +17,7 @@ default = ["flate2", "aws-lc-rs", "rsa"]
 # Crypto backends: enable at least one of `aws-lc-rs` or `ring`.
 aws-lc-rs = ["dep:aws-lc-rs"] # Default crypto backend.
 async-trait = ["dep:async-trait"]
-legacy-ed25519-pkcs8-parser = ["yasna"]
+legacy-ed25519-pkcs8-parser = ["yasna", "num-bigint-04"]
 # Danger: 3DES cipher is insecure.
 des = ["dep:des"]
 # Danger: DSA algorithm is insecure.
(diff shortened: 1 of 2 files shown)


Eugeny:
Fixed indepdendently on main - thanks!

PR #773, opened 2026-09-12
bug: Fix partial_success clobber

### LINKED ISSUE #772: Partial Success Handler Flag Clobbered
Hi! Firstly, thank you for the project! I am grateful for your work. I am building a tiny SSH reverse proxy on top of `russh` and as part of my fuzzy Claude-driven testing of my own tool, I've found that MFA auth doesn't appear to work as expected.

I would like to preface: I am by no means an expert in SSH. I've tried my best to understand the code-path, but the bug was discovered with some heavy Claude usage. The description here is fully hand-written (diagrams and all), and my understanding is my own. Hopefully I've done enough homework to defend that this is a bug. **If you do not think this is a bug, I would love to learn more if you have time, but feel free to close the issue otherwise.** Hopefully I'll have some time to open a PR in a few minutes.

Alright with the preface out of the way, it seems that the MFA flow **rejects** requests on the partial success code-path. My understanding of the MFA path is that a client sends repeated requests to a server, and the server can then respond with a "failure" code, however, if the "failure" is considered to be "partially successful", the server [ought to (rfc4252 5.1)](https://www.rfc-editor.org/info/rfc4252/#section-5.1) send a `[SSH_MSG_USERAUTH_FAILURE, "partial success": true]` response. The client can then provide more data, instead of accepting that the authentication  has failed.

```mermaid
sequenceDiagram
  actor client
  participant server

  note over client, server: first request in the mfa sequence
  client ->> server: [SSH_MSG_USERAUTH_REQUEST, "method-name": "foo", ...] (sec 5)
  server ->> client: [SSH_MSG_USERAUTH_FAILURE, "partial success": true] (sec 5.1)

  note over client, server: next authentication mechanism provided
  client ->> server: [SSH_MSG_USERAUTH_REQUEST, "method-name": "bar", ...]  (sec 5)
  server ->> client: [SSH_MSG_USERAUTH_SUCCESS]
  note over client, server: authentication complete
```

It seems to me that in `Encrypted::server_read_auth_request`, many code-paths take the following structure:

```rust
                let auth = if change {
                    Auth::Reject {
                        proceed_with_methods: None,
                        partial_success: false,
                    }
                } else {
                    handler.auth_password(&user, &password).await?
                };

                if let Auth::Accept = auth {
					... // happy path, looks correct.
                } else {
                    auth_user.clear();
                    if let Auth::Reject {
                        proceed_with_methods: Some(proceed_with_methods),
                        partial_success,
                    } = auth
                    {
                        auth_request.methods = proceed_with_methods;
                        auth_request.partial_success = partial_success;
                    } else {
                        auth_request.methods.remove(MethodKind::Password);
                    }
                    auth_request.partial_success = false;
                    reject_auth_request(until, &mut self.write, auth_request).await?;
                }

```

The last `auth_request.partial_success` seems to clobber the `partial_request` flag that the handler returned, disrespecting the handler's intent.

It seems that the issue was introduced in 0c722b8, perhaps as an oversight. The commit added support for `partial_success`, but appears to have forgotten to remove the explicit `partial_success = false` statements. Note that the `keyboard-interactive` auth mode is not impacted. All other modes seem impacted.

I'll open up a PR in a second with some tests and a fix! Thank you for your work!

markovejnovic:
## Description

- Fixes #772.
- The majority of the diff is in the tests.
  * I wanted to use [`#[rstest]`](https://rstest.rs/) for parametrizing, but there was no precedent.
  * Tried my best to ensure everything is in-line with the existing style.

### Results

All tests pass.

image

- Clippy seems to be failing CI due to existing issues (new clippy added new default rule, I'm guessing).
- Semver seems to be failing due to unrelated things.

### Review Guide

- 9a102c18640bc1584b5dde60b8d43f6b1dc843e5 adds tests. They fail.
image

- 75377be504e56594f78068cd8bc2555054f6fe1f fixes the bug, which causes the tests to succeed.
image

### Other

AI note: The fix is human-designed, human-coded. The tests are human-designed, AI-coded.

## AI Usage

* [ ] Fully vibe coded


markovejnovic:
@Eugeny I am closing this PR for the time being -- I had claude fuzz the change and it found some issues. Will re-open when I investigate.

markovejnovic:
Ah, okay, @eugeny's commit fixes what Claude found. Seems that the issue was that the `partial_success` was mutable, rather than per-request.

Sorry for the close on the PR -- just wanted to make sure we don't add a security regression accidentally and wanted some time to review.

PR #775, opened 2026-09-13
fix: auth-agent-req@openssh.com replies on channel (#774)

### LINKED ISSUE #774: auth-agent-req@openssh.com requests reply on the connection rather than the channel
Similar to my other issue #772, this has been discovered as part of Claude fuzzing an application I'm building on top of `russh`. Similar to that issue, it's important to preface I don't completely understand SSH, so, similar to that issue, feel free to close this issue if it's garbage. I've tried my best to explore the problem domain as best I could (without AI).

It seems that `auth-agent-req@openssh.com` transactions are intended to be channel-scoped, meaning that a channel-scoped request ought to reply on the same channel. Looking at the `OpenSSH` code:

```c
static int
server_input_channel_req(int type, uint32_t seq, struct ssh *ssh)
{
	// ...
	} else if ((c->type == SSH_CHANNEL_LARVAL ||
	    c->type == SSH_CHANNEL_OPEN) && strcmp(c->ctype, "session") == 0)
		success = session_input_channel_req(ssh, c, rtype);

	if (want_reply && !(c->flags & CHAN_CLOSE_SENT)) {
		if (!c->have_remote_id)
			fatal_f("channel %d: no remote_id", c->self);
		if ((r = sshpkt_start(ssh, success ?
		    SSH2_MSG_CHANNEL_SUCCESS : SSH2_MSG_CHANNEL_FAILURE)) != 0 ||
		    (r = sshpkt_put_u32(ssh, c->remote_id)) != 0 ||
		    (r = sshpkt_send(ssh)) != 0)
			sshpkt_fatal(ssh, r, "%s: send reply", __func__);
	}

	// ...
}

int
session_input_channel_req(struct ssh *ssh, Channel *c, const char *rtype)
{
	// ...
	if (c->type == SSH_CHANNEL_LARVAL) {
		// ...
		} else if (strcmp(rtype, "auth-agent-req@openssh.com") == 0) {
			success = session_auth_agent_req(ssh, s, 0);
        }
        // ...
	}
    // ...
}
```

It seems like the handling of `"auth-agent-req@openssh.com"` is in `session.c:session_input_channel_req(struct ssh*, Channel*, const char*)` which is invoked by the `serverloop.c:server_input_channel_req` function, which does this response (on success):

```c
	if (want_reply && !(c->flags & CHAN_CLOSE_SENT)) {
		if (!c->have_remote_id)
			fatal_f("channel %d: no remote_id", c->self);
		if ((r = sshpkt_start(ssh, success ?
		    SSH2_MSG_CHANNEL_SUCCESS : SSH2_MSG_CHANNEL_FAILURE)) != 0 ||
		    (r = sshpkt_put_u32(ssh, c->remote_id)) != 0 ||
		    (r = sshpkt_send(ssh)) != 0)
			sshpkt_fatal(ssh, r, "%s: send reply", __func__);
	}
```

The `SSH2_MSG_CHANNEL_SUCCESS` here appears to be a direct translation of `SSH_MSG_CHANNEL_SUCCESS` as defined in [RFC 4254](https://www.rfc-editor.org/info/rfc4254/#section-5.4):

> ```
>   Many 'channel type' values have extensions that are specific to that
>   particular 'channel type'.  An example is requesting a pty (pseudo
>   terminal) for an interactive session.
>
>   All channel-specific requests use the following format.
>
>      byte      SSH_MSG_CHANNEL_REQUEST
>      uint32    recipient channel
>      string    request type in US-ASCII characters only
>      boolean   want reply
>      ....      type-specific data follows
>
>   (page break)
>
>   If 'want reply' is FALSE, no response will be sent to the request.
>   Otherwise, the recipient responds with either
>   SSH_MSG_CHANNEL_SUCCESS, SSH_MSG_CHANNEL_FAILURE, or request-specific
>   continuation messages.  If the request is not recognized or is not
>   supported for the channel, SSH_MSG_CHANNEL_FAILURE is returned.
> ```

The current logic in `russh` has the following handler in `encrypted.rs`:

```rust
"auth-agent-req@openssh.com" => {
    map_err!(ensure_end(r))?;
    if let Some(chan) = self.channels.get(&channel_num) {
        let _ = chan
            .send(ChannelMsg::AgentForward { want_reply: true })
            .await;
    }
    debug!("handler.agent_request {channel_num:?}");

    let response = handler.agent_request(channel_num, self).await?;
    if response {
        self.request_success()
    } else {
        self.request_failure()
    }
    Ok(())
}
```

and the two response handlers are defined in `server/session.rs`:

```rust
pub fn request_success(&mut self) {
    if self.common.wants_reply {
        if let Some(ref mut enc) = self.common.encrypted {
            self.common.wants_reply = false;
            push_packet!(enc.write, enc.write.push(msg::REQUEST_SUCCESS))
        }
    }
}

pub fn request_failure(&mut self) {
    if let Some(ref mut enc) = self.common.encrypted {
        self.common.wants_reply = false;
        push_packet!(enc.write, enc.write.push(msg::REQUEST_FAILURE))
    }
}
```

Therefore, it seems that `russh` replies with `REQUEST_SUCCESS` as opposed to `CHANNEL_SUCCESS`.

The fix seems easy enough and I'll put a PR up.

markovejnovic:
## Description

- Fixes #774 
- The fix is hand-coded.
- The tests are AI-generated (but heavily human-driven since Claude had bad ideas).
  * I made the tests compare behavior with `OpenSSH` rather than exist in a vacuum.
    - Thought it was wiser since OpenSSH is the reference.
    - Downsides are:
      * You need to install OpenSSH in CI (added in this PR)
      * The test takes a little bit longer than without it.
      
 
### Validation

Confirmed test fails on the test-only commit:

image

Confirmed test passes on the `HEAD` commit:

image

Got claude to review this code in depth

## AI Usage

* [ ] Fully vibe coded


Eugeny:
LGTM - thanks!

Eugeny:
@all-contributors add @markovejnovic for code

